PR #7, opened 2019-04-06
Handling TUF repositories, metadata and targets

renatav:
- Added a new module called `tuf_utils`, containing a class, a context manager and some functions which make it easier to load a TUF repository, add targets and write metadata. TUF expects a directory called `metadata.staged`. `load_repository` loads metadata from that directory. Additionally, TUF repository writes metadata files to the staged directory. Once the metadata is ready, someone has the obligation to move the metadata files to the `metadata` folder. The idea behind the context manager is to automate this step in the following way:
1. `metadata.staged` is created and all metadata files are copied from `metadata` to it.
1. TUF's repository is loaded (an instance of TUF's repository class is created)
1. Instance of the newly added class`TUFRepository` is created and yielded. This class contains methods for adding target files (files which store commits), handling capstone and writing metadata files. I found it more convenient to create a class, as it eliminates to need to pass TUF's repository as a parameter all the time and to construct targets or metadata paths.
1. Copies metadata files from `metadata.stage` to `metadata` and deletes `metadata.staged` since we do not want to commit it.

`tuf_utils` also contains a function for loading keys from keystore files which can, but don't have to be password protected, a helper function for marking a tuf role as dirty.
-  Also, moved all code into a `taf` package. Haven't updated anything inside the `updater` package, so that code should not be reviewed (it's also not ready for a review)
- Created the concept of `repositoriesdb`. This module contains a dictionary of instantiated repositories, as well as functions which can be used to create and load repositories. Repositories are created by reading `repositories.json` and `targets.json` files  at the specified revision, given an atuhentication repo, authentication repository commit and either classes of repositories which should be instantiated or a factory function. This module also has various functions for getting previously loaded repositories.
- Implemented validation of branches:
  - Checking lengths of branches
  - Checking if commit shas of target repositories match commit shas specified in the authentication repository
  - Checking if all commits on a branch have the same branch id
  - Checking if capstone is at the end of the authentication repository's branch
- Updated documentation


taf/oll/targetsdb.py
```diff
@@ -0,0 +1,149 @@
...
+#             'target_path1': (target_git_repository1, custom_data1)
+#             'target_path2': (target_git_repository2, custom_data2)
+#             ...
+#         }
+#     }
+# }
+
+_targetsdb_dict = {}
+
+def load_targets(auth_repo, targets_classes=None, factory=None, root_dir=None, commit=None):
```
danixeee: I think we should not force a user to have a directory structure based on `targets` file.
It's fine to do that by default, but I suggest having another parameter (e.g. `path_mappings`), where we can pass something like:

```python
{
   'cityofsanmateo/law-xml': '/Users/daniel/xml-repos/smc-law',
   'cityofsanmateo/law-html': '/Users/daniel/html-repos/smc-law'
}
```

What are your thoughts?

renatav: We have discussed and solved this. A factory function can be provided if using the default paths is not an option.

taf/oll/cli.py
```diff
@@ -0,0 +1,20 @@
+import click
+from taf.oll.tuf_utils import load_repository
+
+
+@click.group()
+def cli():
+    pass
```
danixeee: Identation should be 2 spaces.

taf/oll/exceptions.py
```diff
@@ -0,0 +1,11 @@
+class InvalidCommit(Exception):
+    pass
```
danixeee: Identation should be 2 spaces. Also, line 5, 8, 11.

taf/oll/tuf_utils.py
```diff
@@ -0,0 +1,221 @@
...
+import json
+import os
+import shutil
+from contextlib import contextmanager
+from pathlib import Path
+import securesystemslib
+
+role_keys_cache = {}
+
+class TUFRepository:
```
danixeee: Maybe it makes more sense to call it `TAFRepository`?

taf/oll/repositoriesdb.py
```diff
@@ -0,0 +1,176 @@
...
+_repositories_dict = {}
+repositories_path = 'targets/repositories.json'
+targets_path = 'metadata/targets.json'
+
+
+def load_repositories(auth_repo, repo_classes=None, factory=None,
+                      root_dir=None, commit=None, only_load_targets=False):
+  """
+  Creates target repositories by reading repositories.json and targets.json files
+  at the specified revision, given an atuhentication repo.
```
danixeee: atuhentication -> authentication

taf/oll/repositoriesdb.py
```diff
@@ -0,0 +1,176 @@
...
+      return False
+
+  paths = list(filter(_compare, repositories)) if custom else list(repositories)
+  if len(paths):
+    return paths
+  raise RepositoriesNotFound(f'Repositories associated with custom data {custom} not found')
+
+
+def get_repository(auth_repo, path, commit=None):
+  return get_repositories(auth_repo, commit)['path']
```
danixeee: I guess it should be `return get_repositories(auth_repo, commit)[path]`.

taf/oll/tuf_utils.py
```diff
@@ -0,0 +1,221 @@
...
+    inside the targets directory.
+    """
+    targets = self._role_obj(targets_role)
+    targets.add_target(os.path.join(self.targets_path, file_path))
+
+
+  def add_targets(self, data, targets_role='targets', files_to_keep=['repositories.json']):
+    """
+    Creates a target .json file containing a repository's commit for each
+    repository. Adds those files to the tuf repostiory. Aslo removes
```
danixeee: repostiory -> repository
aslo -> also

taf/oll/validation.py
```diff
@@ -0,0 +1,111 @@
+import os
+import taf.oll.repositoriesdb as repositoriesdb
```
danixeee: Unused import

taf/oll/validation.py
```diff
@@ -0,0 +1,111 @@
...
+from taf.oll.exceptions import InvalidBranch
+
+def validate_branch(auth_repo, target_repos, branch_name, default_branch='master'):
```
danixeee: `default_branch` param is not used

taf/oll/GitRepository.py
```diff
@@ -0,0 +1,59 @@
+import os
+import json
+from .utils import run
+
+
+class GitRepository(object):
```
danixeee: I would call this class `TAFTarget`, because it represents a target from `targets.json` file, with other information like `additional_info`, `repo_urls`,...
Then we could have a lazy loaded [GitRepo](https://github.com/openlawlibrary/taf/blob/bee39746066da25423227f3628f4740cdd62fbcb/taf/oll/updater/git.py#L29) property, if someone wants to call a git functions for that target.

renatav: I thought about both ideas before, but I decided to name and implement `GitRepository` like this for a few reasons. We  instantiate the repositories based on `repositories.json`, meaning that TAF can instantiate a repository which is not specified in `targets.json`. This is important to us because we need to handle preview repositories, but I think that it's a generally good idea to support that. Since TAF is designed to handle git repositories, I think that having a class named `GitRepository` makes sense.
I also thought about having a Git repository as a property, but then we would have to call `git_repo.some_git_method` every time instead of just `some_fit_method`. Since the class is supposed to represent a Git repository anyway, I think that it makes sense to put all methods directly in it. I though about `additional_info`, and at first I didn't like storing such data inside `GitRepository`. However, since that additional info is specified in `repositories.json` or `targets.json` and is directly related to the repository, I prefer the final solution.

taf/oll/validation.py
renatav: We expect what everything will be merged into master. So, the idea was to support different default branches. I decided not to implement that now (but forgot to remove the additional parameter), but we can create an issue to generalize this validation as possible if you agree

taf/oll/tuf_utils.py
renatav: I am not sure which name is better. This class is meant to be a wrapper around TUF repository

taf/oll/repositoriesdb.py
renatav: Yes

taf/oll/tuf_utils.py
danixeee: For me it's little bit confusing.
I am just thinking of instantiating this class in following way:

```python
repository = TUFRepository()
tuf_repo = repository.repository # which is `tuf.repository_tool.Repository`
```

Does it makes more sense to subclass `tuf.repository_tool.Repository`?

taf/oll/validation.py
danixeee: I just wanted to point that out because I wasn't sure if you forgot to use that parameter somewhere, or skipped implementing it for now. You can decide what is better for now

taf/oll/GitRepository.py
danixeee: I am fine with that, just this class seemed too empty for me. I would suggest just adding some basic git functionalities to this class then, what do you think?

renatav: I added just what we need in TAF. Do you think it would make sense to add functionalities like commit, push, pull, create branch and such?

taf/oll/tuf_utils.py
renatav: The problem is that `tuf.repository_tool.Repository` is instantiated by TUF, using `load_repository`. While I agree that it would be nice to subclass it, I don't know how to do that elegantly. Do you have any suggestions?

taf/oll/validation.py
renatav: I'll remove the parameter for now

renatav:
@danixeee And regarding the other comments, yes the folder should be removed. I removed it locally but it seems like I did not commit that change at any point. I have not taken a look at the updater in a while. I just copy-pasted platform's `GitRepo` class as it was when I started working on the updater. I did not put much thought into it. I would leave updater as it is for now, and refactor everything later. That `run` function should also be deleted, I agree.
Yes, we do need `oll-tuf` dependency, because we need the changes made to TUF's updater class. We will update that once the TUF team implements that separation of responsibilities that we need.

taf/oll/GitRepository.py
danixeee: I think yes, and if someone wants some specific methods, custom subclass can be provided. But for me it makes sense to have at least `git pull`.

taf/oll/validation.py
renatav: Created an issue for making validation more generic

taf/oll/GitRepository.py
renatav: I added moved a few more methods from platform. I think that, since TAF has functionalities for validating branches, it makes sense to add functionalities for creating and checking out branches. Also added basics like commit, push and pull

renatav:
@danixeee  I took another look at the folder containing the udpater you mentioned. I did remove the old folder, so that all seems fine to me.

taf/oll/tuf_utils.py
danixeee: We also use `taf.tuf_utils.load_repository` to instantiate `taf.tuf_utils.TUFRepository`, right?
I guess it should work the same way if you just subclass `tuf.repository_tool.Repository` and just use `self.SOME_METHOD_FROM_SUBCLASS` instead of `self.repository.SOME_METHOD_FROM_SUBCLASS`?

Since we are just extending a TUF that way, we can maybe use the same module and class names.

```python
import taf.repository_tool.load_repository

with load_repository() as repo:
   repo...
```

setup.py
```diff
@@ -38,6 +38,11 @@
   install_requires = [
     'oll-tuf'
   ],
+  entry_points={
+        'console_scripts': [
+            'taf = taf.oll.cli:main'
```
danixeee: Should be `taf.cli:main`.

taf/cli.py
```diff
@@ -0,0 +1,21 @@
...
+def cli():
+  pass
+
+
+@cli.command()
+@click.option('--repo-path', help='Authentication repository\'s path')
+@click.option('--file-path', help='Target file\'s path, relative to the targets directory')
+@click.option('--keystore-path', help='Path of the keystore file')
+@click.option('--targets-role', default='targets', help='Targets metadata file to be updated')
+def add_target_file(repo_path, file_path, keystore_path, targets_role):
```
danixeee: Should we make first three click options required here? Or we can use `click.argument` for those.

setup.py
danixeee: Line 2. in the same file (setup.py): unused import

taf/repository_tool.py
```diff
@@ -0,0 +1,221 @@
...
+  def add_existing_target(self, file_path, targets_role='targets'):
+    """
+    Registers new target files with TUF. The files are expected to be
+    inside the targets directory.
+    """
+    targets = self._role_obj(targets_role)
+    targets.add_target(os.path.join(self.targets_path, file_path))
+
+
+  def add_targets(self, data, targets_role='targets', files_to_keep=['repositories.json']):
```
danixeee: File `repositories.json` should always exist (used in _repositoriesdb.py_), so, in my opinion `files_to_keep` should be a constant or field in `Repository` class. E.G. Passing an empty array will break something probably, right?

taf/cli.py
renatav: The first three parameters are required. Yes, I think it makes sense to use `argument`

taf/repository_tool.py
renatav: Yes, `repositories.json` should always exist. We need to consider two things though. There might be a need to update `repositories.json`, in which case we don't want to keep the old version. Secondly, someone using TAF might want to handle another file in a similar way. I guess that we could have a field which lists all files required by the framework, and still allow the users to specify additional files when calling `add_targets`, depending on their needs at a particular time. If `repositories.json` is also in `data`, we won't copy the previous version

danixeee: Maybe we need more methods to work with `repositories.json` file since it is a part of framework (in the future)? I can't think of example where user would want another "special" target file because we already have `additional_info` field to store custom data.

renatav: I think that it's a good idea to keep this flexible. files_to_keep should contain a list of all files which should remain targets, but should not be modified. Meaning that there is no need to read their content and pass it as an item in the data dictionary. I thought that this option would be useful. The targets do not even have to have any particular importance.
I like the idea of adding methods for working with repositories.json. I'll create an issue

taf/repository_tool.py
```diff
@@ -0,0 +1,233 @@
...
+
+  def add_existing_target(self, file_path, targets_role='targets'):
+    """
+    Registers new target files with TUF. The files are expected to be
+    inside the targets directory.
+    """
+    targets = self._role_obj(targets_role)
+    targets.add_target(os.path.join(self.targets_path, file_path))
+
+  def add_targets(self, data, targets_role='targets', files_to_keep=[]):
```
danixeee: Please change `files_to_keep` to `None` default value instead of `[]`, because the _same_ default list will be used for every `add_targets` call, if `files_to_keep` argument is not passed.

```python
required_files = ['test']

def test(files_to_keep=[]):
  if files_to_keep is None:
    files_to_keep = []
  files_to_keep.extend(required_files)
  print(files_to_keep)

test()
test()
```

Output is:

```
['test']
['test', 'test']
```

renatav: Good catch. I forgot about that. Fixed

PR #11, opened 2019-05-08
Some repositoriesdb improvements

renatav:
- Raising a nicer error if repositories.json or mirrors.json does not exist or is not valid
- Added the ability to pass in more than one commit when loading repositories
- Finding a deduplicated list of repositories given one or more commits

### DIFF
--- a/taf/GitRepository.py
+++ b/taf/GitRepository.py
@@ -1,8 +1,8 @@
 import json
 import os
 import subprocess
-
-from .utils import run
+from pathlib import Path
+from taf.utils import run
 
 
 class GitRepository(object):
@@ -20,11 +20,7 @@ class GitRepository(object):
     root_dir.
     """
     self.target_path = target_path
-    if target_path is not None:
-      self.repo_path = os.path.join(root_dir, target_path.replace('/', os.sep))
-    else:
-      self.repo_path = root_dir
-    self.repo_path = os.path.normpath(self.repo_path)
+    self.repo_path = str((Path(root_dir) / (target_path or '')).resolve(True))
     self.repo_urls = repo_urls
     self.additional_info = additional_info
 
--- a/taf/GitRepository.py
+++ b/taf/GitRepository.py
@@ -24,6 +24,7 @@ class GitRepository(object):
       self.repo_path = os.path.join(root_dir, target_path.replace('/', os.sep))
     else:
       self.repo_path = root_dir
+    self.repo_path = os.path.normpath(self.repo_path)
     self.repo_urls = repo_urls
     self.additional_info = additional_info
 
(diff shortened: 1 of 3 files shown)


taf/repositoriesdb.py
```diff
@@ -54,37 +55,43 @@ def load_repositories(auth_repo, repo_classes=None, factory=None,
...
-      raise Exception(f'{type(git_repo)} is not a subclass of GitRepository')
+      if not isinstance(git_repo, GitRepository):
+        raise Exception(f'{type(git_repo)} is not a subclass of GitRepository')
 
-    repositories_dict[path] = git_repo
+      repositories_dict[path] = git_repo
+  get_deduplicated_repositories(auth_repo, commits[0])
 
 
 def _determine_repo_class(repo_classes, path):
```
danixeee: I know this is not changed, but I missed it in the last CR. This should work the same:

```python
def _determine_repo_class(repo_classes, path='default'):
  # if no class is specified, return the default one
  if repo_classes is None:
    return GitRepository
  # if only one value is specified, that means that all target repositories
  # should be of the same class
  if not isinstance(repo_classes, dict):
    return repo_classes

  return repo_classes.get(path, GitRepository)
```

taf/GitRepository.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, root_dir, target_path=None, repo_urls=None, additional_info=N
       self.repo_path = os.path.join(root_dir, target_path.replace('/', os.sep))
```
danixeee: Maybe we can use `pathlib` like in `repositoriesdb.py` module?

taf/repositoriesdb.py
```diff
@@ -54,37 +55,43 @@ def load_repositories(auth_repo, repo_classes=None, factory=None,
   elif commit in _repositories_dict[auth_repo.name]:
```
danixeee: `commit` variable does not exist anymore

taf/repositoriesdb.py
```diff
@@ -19,7 +20,7 @@
 
 
 def load_repositories(auth_repo, repo_classes=None, factory=None,
-                      root_dir=None, commit=None, only_load_targets=False):
+                      root_dir=None, only_load_targets=False, *commits):
   """
   Creates target repositories by reading repositories.json and targets.json files
```
danixeee: Old docstring (line 45.)

taf/GitRepository.py
renatav: I would prefer not to modify target path because it is specified in both `repositories.json` and `mirrors.json` with `/`. I think it's better to keep it consistent with that

danixeee: Sure, but I still think that we can simplify it.

```python
self.target_path =  target_path
self.repo_path = (Path(root_dir) / (target_path or '').replace('/', os.sep)).resolve(True)
```

I think `pathlib` handler os separators, so maybe we can remove `.replace('/', os.sep)`

PR #13, opened 2019-05-15
Authentication repository updater

renatav:
- Refactored the initial updater, so that it uses the `GitRepository` class.
- Renamed the handler class, since it can be used to download both metadata files and targets. We no longer specify commits as targets directly, so a separate targets handler is not necessary.
- Finished the initial version of the authentication repository's updater. Target repositories are validated and updated as well.
- Added validation of the names and urls of repositories
- Added cli functions which made development easier, but seem generally useful. They can be used to add targets, set expiration dates of the metadata files and call the update process.


taf/GitRepository.py
```diff
@@ -44,6 +57,30 @@ def checkout_branch(self, branch_name, create=False):
       else:
         raise(e)
 
+  def clone_or_pull_up_to_commit(self, commit):
+    if not self.is_git_repository():
+      self.init_repo()
+    self.fetch()
+    self.merge_commit(commit)
+
+  def clone(self, no_checkout=False):
```
danixeee: I think we shuld have another param (e.g. `recreate`) to _explicitly_ tell that directory at `self.repo_path` will be deleted and recreated.

Moreover, `AuthenticationRepo` is a subclass of `GitRepository` and we don't want to accidentally remove the existing repo on a file system (I know that it will be cloned only if it does not exist, but anyway).

I think `repo.clone(True)` to remove directory and clone repo again is little bit safer.

taf/GitRepository.py
```diff
@@ -44,6 +57,30 @@ def checkout_branch(self, branch_name, create=False):
...
+    shutil.rmtree(self.repo_path, True)
+    os.makedirs(self.repo_path, exist_ok=True)
+    if self.repo_urls is None:
+      raise Exception('Cannot clone repository. No urls were specified')
+    params = ''
+    if self.bare:
+      params = '--bare'
+    elif no_checkout:
+      params == '--no-checkout'
```
danixeee: `==` -> `=`

taf/GitRepository.py
```diff
@@ -81,7 +122,22 @@ def get_file(self, commit, path):
 
   def head_commit_sha(self):
     """Finds sha of the commit to which the current HEAD points"""
-    return self._git('rev-parse HEAD')
+    try:
+      return self._git('rev-parse HEAD')
+    except:
```
danixeee: We should catch `CalledProcessError`

taf/cli.py
```diff
@@ -18,5 +19,40 @@ def add_target_file(repo_path, file_path, keystore_path, targets_role):
...
+    if expiration_date is not None:
+      repository.set_metadata_expiration_date(metadata_role, expiration_date, 0)
+    else:
+      repository.set_metadata_expiration_date(metadata_role)
+    repository.write_roles_metadata(metadata_role, keystore_path)
+
+
+@cli.command()
+@click.option('--url', help='Authentication repository\'s url')
+@click.option('--clients-directory', help='Directory containing the client\'s authentication repository')
```
danixeee: I found this option too long, I think `--clients-dir` is slightly better. Also, for help arguments, you can avoid escaping `'` if you use double quotes `help="Authentication repository's url"`

taf/cli.py
```diff
@@ -18,5 +19,40 @@ def add_target_file(repo_path, file_path, keystore_path, targets_role):
     tuf_repo.add_existing_target(file_path)
     tuf_repo.write_roles_metadata(targets_role, keystore_path, True)
 
+@cli.command()
+@click.option('--repo-path', help='Authentication repository\'s path')
+@click.option('--keystore-path', help='Path of the keystore file')
+@click.option('--targets-role', default='targets', help='Targets metadata file to be updated')
+def add_targets(repo_path, keystore_path, targets_role):
+  targets_path = os.path.join(repo_path, 'targets')
```
danixeee: `CLI` should be just a wrapper around our core logic. We should wrap below lines in a core function and call it in one line (like you did for `taf update` CLI).
I can change this once I start working on TAF Management app.

taf/updater/handlers.py
```diff
@@ -68,82 +84,170 @@ def __init__(self, mirrors, repository_directory):
...
     # next commit as current for the given file
-    #last_index = len(self.commits) - 1
-    self.commits_indexes[filename] = self.commits.index(location)
-
-  def cleanup(self):
-    # this needs to be called after the update is finished
-    # either successfully or unsuccessfully
-    shutil.rmtree(self.current_path)
-    shutil.rmtree(self.previous_path)
+    print(f'{filename} updated from {location}')
```
danixeee: Parameter `location` should be `mirror` (same as in base class)

taf/updater/handlers.py
```diff
@@ -68,82 +84,170 @@ def __init__(self, mirrors, repository_directory):
...
-  def get_mirrors(self, remote_filename):
+  def cleanup(self):
+    """
+    Removes the bare authentication repositoy and current and previous
+    directories. This should be called after the update is finished,
+    either successfully or unsuccessfully.
+    """
+    shutil.rmtree(self.current_path)
+    shutil.rmtree(self.previous_path)
+    os.system(f'rmdir /S /Q "{self.validation_auth_repo.repo_path}"')
```
danixeee: Why you didn't use `shutil.rmtree` here as well?

taf/updater/handlers.py
```diff
@@ -68,82 +84,170 @@ def __init__(self, mirrors, repository_directory):
...
+
+  def earliest_valid_expiration_time(self, metadata_rolename):
```
danixeee: Parameter `metadata_rolename` is not used here nor in baseclass, should we remove it then?

taf/updater/updater.py
```diff
@@ -58,38 +59,49 @@ def update():
...
+  # TODO old HEAD as an input parameter
 
-  clients_reposiotry = os.path.join(clients_directory, repository_name)
-  clients_keystore = os.path.join(clients_directory, 'keystore')
-  clients_metadata = os.path.join(clients_reposiotry, 'metadata')
+  clients_repository = os.path.join(clients_directory, repo_name)
 
   # Setting 'tuf.settings.repository_directory' with the temporary client
   # directory copied from the original repository files.
   tuf.settings.repositories_directory = clients_directory
```
danixeee: `clients_repository` from line 64. is not used. Did you mean `tuf.settings.repositories_directory = clients_repository`?

taf/GitRepository.py
renatav: I added ` shutil.rmtree(self.repo_path, True)` to ease development. It should definitely be removed before this is merged.

taf/cli.py
renatav: I personally prefer escpaing string than starting some string with " and some with '. You're probably right though, using "". Sure, `clients-dir` sounds fine

taf/cli.py
renatav: I agree. We could do something similar to how we handle build in platform. This was just something I quickly put together in order to be able to generate initial auth repository.

taf/updater/handlers.py
renatav: It didn't work. Was getting a permission error

taf/updater/handlers.py
renatav: Yes. I thought that I would need it at first and forgot to remove it.

taf/updater/updater.py
renatav: I think that so, but it doesn't seem to make any difference.

renatav:
@danixeee This should still execute all of TUF's checks, plus one additional one (checking it the file remained the same). If you could set up automated checks like you did in pygls, that would be great.

taf/updater/handlers.py
danixeee: What I can see from code is that you are using `tempfile` to create a temporary directory [here](https://github.com/openlawlibrary/taf/blob/renatav/updater/taf/updater/handlers.py#L185). 

```python
>>> import tempfile
>>> print(tempfile.gettempdir())
/var/folders/lc/c_lvyn5x6cbd12ysjly61qcr0000gn/T
```

This is what I get on my machine (macOS).
I suggest adding another child directory, with a random name, like: `/var/folders/lc/c_lvyn5x6cbd12ysjly61qcr0000gn/T/awled12k34n1`, to avoid possible conflicts.

> It didn't work. Was getting a permission error

Can you try this:

```python
temp_dir = tempfile.mkdtemp() # does what I mentioned above; instead of tempfile.gettempdir()
self.validation_auth_repo = AuthenticationRepo(temp_dir, self.metadata_path, self.targets_path, repo_name, [url], True)

shutil.rmtree(self.validation_auth_repo.repo_path)
```

@renatav I will push it on [my branch](https://github.com/openlawlibrary/taf/blob/danixeee/setup-formatter-and-bandit) where I fixed some `pylint` and `bandit` issues for the updater as well. Please test when you have time.

renatav: I liked cloning the repository directly inside temp, as that made testing easier (it's easier to find it inside temp, see what it looks like, delete it manually if needed etc.). Update of the authentication repository seems stable now, so I will look into fixing this. I'll test if your method works on Windows

setup.py
```diff
@@ -13,45 +13,68 @@
...
+    ],
+    zip_safe=False,
+    install_requires=[
+        'click==6.7',
+        'cryptography==2.3.1',
+        'oll-tuf==0.11.2.dev5',
+    ],
+    extras_require={
+        'ci': ci_require,
+        'test': tests_require,
```
danixeee: Please add `'dev': dev_require,` here (I forgot).

taf/GitRepository.py
```diff
@@ -20,32 +22,85 @@ def __init__(self, root_dir, target_path=None, repo_urls=None, additional_info=N
...
     self.target_path = target_path
-    self.repo_path = str((Path(root_dir) / (target_path or '')).resolve(True))
+    self.repo_path = str((Path(root_dir) / (target_path or '')).resolve())
     self.repo_urls = repo_urls
     self.additional_info = additional_info
+    self.bare = bare
+
+  @property
+  def is_git_repository(self):
+    return self._is_git_repository()
```
danixeee: You can remove `_is_git_repository` method and implement it in this property.

taf/GitRepository.py
```diff
@@ -20,32 +22,85 @@ def __init__(self, root_dir, target_path=None, repo_urls=None, additional_info=N
...
+    else:
+      commits = self._git('log --format=format:%H').strip()
+    if not commits:
+      return []
+    commits = commits.split('\n')
+    commits.reverse()
+    return commits
+
+  def all_fetched_commits(self, branch='master'):
+    commits = self._git('rev-list ..origin/{}'.format(branch)).strip()
```
danixeee: We can now call `_git` method this way: `self._git('rev-list ..origin/{}', branch).strip()`.

taf/updater/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class UpdateFailed(Exception):
```
danixeee: It is a convention to add `Error` suffix if exception is actually an error.

taf/GitRepository.py
```diff
@@ -20,32 +22,85 @@ def __init__(self, root_dir, target_path=None, repo_urls=None, additional_info=N
...
+      self._git('checkout {}', branch_name)
     except subprocess.CalledProcessError as e:
       if create:
         self.create_and_checkout_branch(branch_name)
       else:
         raise(e)
 
+  def clone(self, no_checkout=False, from_filesystem=True):
+    # TODO this creates something like smc-law/smc-law instead of just smc-law
+    shutil.rmtree(self.repo_path, True)
```
danixeee: Do you plan to remove this now or to add another parameter to indicate that existing repo will be removed from local and cloned again?

taf/cli.py
```diff
@@ -19,4 +22,44 @@ def add_target_file(repo_path, file_path, keystore_path, targets_role):
...
+    if expiration_date is not None:
+      expiration_date = datetime.datetime.strptime(expiration_date, '%Y-%m-%d')
+      repository.set_metadata_expiration_date(metadata_role, expiration_date, 0)
+    else:
+      repository.set_metadata_expiration_date(metadata_role)
+    repository.write_roles_metadata(metadata_role, keystore_path)
+
+
+@cli.command()
+@click.option('--url', default='E:\\OLL2\\updater\\smc-law', help="Authentication repository's url")
```
danixeee: Please remove default values before merging.

taf/updater/exceptions.py
danixeee: Also, there is already `taf.exceptions` module, can you move this exception to it? Can you add `Error` suffix to all of them, as well?
It is a good practice to have base exception like `TAFError` and to subclass it for all specific TAF errors.
Like here: https://github.com/openlawlibrary/taf-sc/blob/danixeee/api-tests/taf_sc/exceptions.py#L5

renatav: Yeah, I've done so in the past, but it slipped my mind. Will update.

taf/GitRepository.py
renatav: I would like to integrate the updater with the publish server before merging, as there will surely be many more changes. Will go through the code and clean it up before this is merging.

taf/cli.py
renatav: Will do, but I would not merge this just yet.

taf/updater/updater.py
renatav: It did make a difference. This is resolved though.

taf/updater/exceptions.py
renatav: Done

taf/GitRepository.py
renatav: Let's leave this for now. Clone will not be called if the repository already exists. Let's leave the TODO as well and give it another thought later.

taf/cli.py
```diff
@@ -19,4 +22,46 @@ def add_target_file(repo_path, file_path, keystore_path, targets_role):
...
+    else:
+      repository.set_metadata_expiration_date(metadata_role)
+    repository.write_roles_metadata(metadata_role, keystore_path)
+
+
+@cli.command()
+@click.option('--url', help="Authentication repository's url")
+@click.option('--clients-dir', help="Directory containing the client's authentication repository")
+@click.option('--repo-name', help="Authentication repository's name")
+@click.option('--targets_dir', help="Directory containing the target repositories")
```
danixeee: `--targets_dir` -> `--targets-dir`
`--from-filesystem` -> `--from-fs` maybe?

danixeee: And `--from-fs` could be a _flag_ option

PR #24, opened 2019-06-05
Setting repository name and path fixes

renatav:
Removed `repo_name` input parameter and its validation from `GitRepository` class and created a `NamedGitRepositoty` class which expects that parameter and validates it. Updater can also be invoked with and without providing that parameter. If it is not provided, it is necessary to provide the repository's full path. Also updated cli, so that both update functions can be invoked.


renatav:
@danixeee The `name` property is not longer needed. The idea is that every repository will now have a `repo_name` attribute, which will either be provided during instantiation of the repository or calculated based on the required `root_dir` parameter. If only `root_dir` is provided, it is expected that it is the repo's full path. It could be more intuitive if we had separate `root_dir` and `repo_path` input parameters, although this would mean that we would have to add a check and raise an error if none of them is provided. So, the code would be a bit more complex. I am not sure what's the best thing to do here. I just realized that I need to update the doc string if this is left as is.
The reason why I decided to rename `targets_path` to `repo_name` is to also support authentication repositories. For target repositories, `repo_name` should correspond to the defined targets paths.
I don't understand what we would gain from moving validation of the repo's name elsewhere. We do not want to restrict validation to target repositories. We do want to validate the name of an authentication repository if it is provided, so the `GitRepostitory` seems to be the best place for this validation.
I don't think that we should generalize the validation to support filesystem paths, but we could add a separate validation. Local paths are only needed for testing, so I don't think that that is a priority. 
About `update_from_filesystem`, I think that it is important to have a distinction between URLs which are files system paths, although they shouldn't be, and intentional file system paths. My thoughts are that file system urls should only be used for testing, so I am not sure abut another input parameter. I have an idea though, will push those changes.
I initially thought that `bare` could play a bigger role than just while cloning. It does not seem like that was a correct assumption. Will update.

renatav:
I took another look at `update_from_filesystem` and the reason why I initially implemented it using the settings module is because of the problems caused by integrating our code with the TUF updater. It TUF, it is common to rely on settings like this, so that seemed like a good solution. Since we will have to do more refactoring once the TUF team is done with their work, I would live this as is for now.

danixeee:
@renatav I still think that it is confusing, because `repo_name` is not actually repository name, it is a targets path. Validation could be bypassed anytime by not passing that argument, e.g.:

```python
root_dir = '/Users/me/repositories'
repo_name = 'myorg/myrepo'

repo = GitRepository(root_dir + repo_name, ...) # same as GitRepository(root_dir, repo_name, ...) but skips validation

# But if we pass this
repo = GitRepository(root_dir, ...)
repo.repo_name # == 'me/repositories'
```

so if `GitRepository` instance could be created with _or_ without validation, I don't see the point of having it at that place. It makes me more sense to validate targets path when loading repositories and creating `GitRepository` instances _for targets_. It also make sense to subclass `GitRepository` and add that validation there.

renatav:
@danixeee  `repo_name` is namespace prefixed name. We could use `targets_path` if we decided to separate authentication and target repositories, but that seems like an unnecessary complication. I would say that `repo_name` with both cases well enough, unlike `targets_path`. Yes, the validation can be avoided, so it is up to the caller to decide if they want the validation or not. The way I see it is that, if you want to validate a name of the repository, you will pass it. We could maybe make the validation flexible by allowing a custom regular expression to be provided. If we moved validation to where we load target repositories, we would then have to have validation in two places, so that we can also validate authentication repository's name when that is needed. I guess that we could have a `GitRepository`, without name validation and another class which would extend it and implement validation. It would still be up to the caller to decide if they want to avoid validation or not. If you **really** don't like that validation, let's just do that and close that issue

danixeee:
> The way I see it is that, if you want to validate a name of the repository, you will pass it.

This is true, but the main problem is that `root_dir` parameter has two meanings, depending on `repo_name`. You could have `repo_path` instead of `root_dir` and just do validation for `repo_name` name if it is passed.

When I see `GitRepository` class, I would expect that `repo_name` is a last directory name in `repo_path`. As it is now, getting _repo name_ for repo location `/Users/me/repos/something/repo-name` will return `something/repo-name` which does not make sense. Now we know what we are doing, but in a few months it will be confusing again.
Having a `TargetsRepository` subclass makes the most sense to me, and you could have `targets_path` there and do the validation. For custom targets classes, user will subclass it.

renatav:
The problem is that we are want to pass and validate authentication repository's name in some cases. So, it would make the least sense to intantiate a TargetRepository class in such a case. I am also not thrilles with root_dir having two meanings, as discussed above, but I also don't think that it really isn't that important. Since we've already put a lot of time into thibking about this, we might as well take some more to implement the celanest solution. I can refactor GitRepository and extend it for the case when we want to split the path into root_dir and repo_name

taf/git.py
```diff
@@ -12,35 +12,30 @@
...
     """
     Args:
-      root_dir: the root directory, repo_name is relative to it
-      repo_name: repository's relative path, as specified in targets.json
-      in case of target repositories (oprional)
+      repo_path: repository's path
       repo_urls: repository's urls (optional)
       additional_info: a dictionary containing other data (optional)
-    repo_path is the absolute path to this repository. If target path is not None,
+    repo_path is the absolute path to this repository. If repo_name is not None,
```
danixeee: _docstring_ should be updated here

taf/cli.py
```diff
@@ -64,12 +63,22 @@ def add_targets(repo_path, targets_key_slot, targets_key_pin, update_all, keysto
 @cli.command()
 @click.option('--url', help="Authentication repository's url")
 @click.option('--clients-dir', help="Directory containing the client's authentication repository")
-@click.option('--repo-name', help="Authentication repository's name")
-@click.option('--targets-dir', help="Directory containing the target repositories")
+@click.option('--targets_dir', help="Directory containing the target repositories")
```
danixeee: `--targets_dir` -> `--targets-dir`

taf/cli.py
```diff
@@ -64,12 +63,22 @@ def add_targets(repo_path, targets_key_slot, targets_key_pin, update_all, keysto
...
-  taf_updater(url, clients_dir, repo_name, targets_dir, from_fs)
+def update(url, clients_dir, targets_dir, from_fs):
+  update_repository(url, clients_dir, targets_dir, from_fs)
+
+
+@cli.command()
+@click.option('--url', help="Authentication repository's url")
+@click.option('--clients-dir', help="Directory containing the client's authentication repository")
+@click.option('--repo-name', hrlp="Repository's name")
+@click.option('--targets_dir', help="Directory containing the target repositories")
```
danixeee: `--targets_dir` -> `--targets-dir`

PR #25, opened 2019-06-07
Logging and saving the last validated commit to a file

renatav:
- Implemented logging and replaced prints with log calls
- Logging can be configured using the settings module. It is possible to log to console, file, or both, and to log errors to a separate file. Currently, errors will be logged to both files, but I think that is fine.
- Added a flag for logging output of commands. I am not sure if we want to include them in our logs, not even at the debug level. I think that they will just clutter up the log. The output can be enabled during development, if it seems like something that can provide useful information.
- Last validated file is saved to a file. It is read during the update process. Added a check which makes sure that the saved commit and the head commit of the authentication repository match. Added some notes about future improvements of this step.
- Some refactoring


taf/git.py
```diff
@@ -33,34 +34,65 @@ def __init__(self, repo_path, repo_urls=None, additional_info=None):
...
+    log_error_msg = kwargs.pop('log_error_msg', '')
+    reraise_error = kwargs.pop('reraise_error', False)
+    log_success_msg = kwargs.pop('log_success_msg', '')
+
     if len(args):
       cmd = cmd.format(*args)
-    return run('git -C {} {}'.format(self.repo_path, cmd))
+    command = 'git -C {} {}'.format(self.repo_path, cmd)
+
+    if log_error or log_error_msg:
```
danixeee: I think passing _logging_ arguments to `_git` method is little bit confusing.

I think you can implement something similar with decorator and use it instead:

```python
def log(success_msg, error_msg, re_raise=False):
    def _log(f):
        def wrapper(self, *args):
            try:
                logger.debug("Repo %s: %s", self.repo_name, success_msg)
                res = f(self, *args)
                logger.debug("After repo %s: %s", self.repo_name, res)
                return res
            except:
                if error_msg:
                    logger.error("Repo %s: %s", self.repo_name, error_msg)
                if re_raise:
                    raise
        return wrapper
    return _log
```

`run` function should always log calls and we will see something like:
`git -c path/to/repo commit -m "Message"`, so maybe that's just enough for now.

danixeee: And use it like this:

```python
@log(
    'Check out branch',
    'Failed to check out branch',
    re_raise=True
)
def checkout_branch(self, branch_name, create=False):
    """..."""
```

Log output e.g.: "Repo taf: Check out branch ['master']".

taf/repositoriesdb.py
```diff
@@ -70,7 +78,7 @@ def load_repositories(auth_repo, repo_classes=None, factory=None,
       repositories = _get_json_file(auth_repo, repositories_path, commit)
       targets = _get_json_file(auth_repo, targets_path, commit)
     except InvalidOrMissingMetadataError as e:
-      print('Skipping commit {}. {}'.format(commit, e))
+      logger.debug('Skipping commit %s due to error %s', commit, e)
```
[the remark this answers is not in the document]
renatav: Info or warning might be better now that I think about it

taf/utils.py
```diff
@@ -11,7 +12,8 @@ def run(*command, **kwargs):
   stdout."""
   if len(command) == 1 and isinstance(command[0], str):
     command = command[0].split()
-  logger.debug('About to run command %s', ' '.join(command))
+  if taf.settings.LOG_COMMAND_OUTPUT:
```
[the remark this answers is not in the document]
renatav: I found looking at logs really hard due to outputs of various git calls. I think that we should make sure to log errors. Outputs of commands can be logged when necessary, in their callers. `git show` really cluttered up the log.

taf/git.py
renatav: There seem to be some libraries for this. See https://www.sighalt.de/remove-visual-noise-of-logging-code-by-using-python-decorators.html

taf/utils.py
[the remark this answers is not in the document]
danixeee: I agree. Does it make sense maybe to log this on debug level (always), and in `git.py` to log on info level (e.g. `Repo my-repo: Checking git status`) ?

taf/git.py
danixeee: I was looking at [this method](https://github.com/openlawlibrary/taf/blob/11776f22f1cdcb2c7db9f4190b646ee827e51d7c/taf/git.py#L70) for example. Looking just at the log file, we won't know if this method is called or some other, because `_git` method will log something like `Repo my-repo: rev-list 123abc..HEAD`, but I might be wrong. Anyway, since it's not priority and it does the job I am fine with that.

taf/updater/auth_repo.py
```diff
@@ -0,0 +1,127 @@
...
+    repositories_commits = defaultdict(list)
+    targets = self.target_commits_at_revisions(commits)
+    previous_commits = {}
+    for commit in commits:
+      for target_path, target_commit in targets[commit].items():
+        previous_commit = previous_commits.get(target_path)
+        if previous_commit is None or target_commit != previous_commit:
+          repositories_commits[target_path].append(target_commit)
+        previous_commits[target_path] = target_commit
+    logger.debug('Auth repo %s: new commits per repositories accoring to targets.json: %s',
```
danixeee: accoring -> according

taf/git.py
renatav: With the current setup, we can add calls to methods which call `_git`. Let's add an issue for improving our logging strategy and address it later, when there aren't as many high-priority tasks.

renatav: Done: https://github.com/openlawlibrary/taf/issues/27

taf/utils.py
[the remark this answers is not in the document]
renatav: I think that we need another logging level, something below debug. I don't think that something like 'Getting HEAD commit' or 'Checking out a branch' should be logged using info. I think that debug is better option in that case. We will use debug during development and testing and I feel like various outputs of git commands will be problematic. I found this [stack overflow](https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility) discussion. What do you think? Is this something we should do now? It does not seem like that would take long.

PR #41, opened 2019-07-06
Fixed update of repositories which allow unauthenticated commits

renatav:
The update used to fail when a repository which allows unauthenticated commits had no new commits or only commits which are not registered in the authentication repository.
Testing the case when no update is necessary and one of the target repositories allows unauthenticated commits.

### DIFF
--- a/taf/updater/updater.py
+++ b/taf/updater/updater.py
@@ -204,9 +204,9 @@ def _update_target_repositories(repositories, repositories_json, repositories_co
     else:
       repository.fetch(True)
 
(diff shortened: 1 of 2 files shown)


taf/updater/updater.py
```diff
@@ -195,8 +204,18 @@ def _update_target_repositories(repositories, repositories_json, repositories_co
     else:
       repository.fetch(True)
 
+    if allow_unauthenticated and old_head is not None:
```
danixeee: You can move first part of this if-statement under below one (`if old_head is not None:`).

PR #42, opened 2019-07-06
Add targets tests and fixes

### LINKED ISSUE #40: `add_targets` does not support relative paths

renatav:
- Add targets no longer deletes repositories defined in `repositories.json`
- Add targets should now support relative paths. Tested on Windows. Please test on another OS.
- Using `pathlib` instead of heavily relying on the `os` module. Conversions to `str` are necessary in order to support Python 3.5.
- Added two add targets tests.

Closes #40


taf/repository_tool.py
```diff
@@ -264,38 +277,41 @@ def add_targets(self, data, targets_role='targets', files_to_keep=None):
       files_to_keep = []
     # leave all files required by the framework and additional files specified by the user
     files_to_keep.extend(self._required_files)
-
+    # add all repositories defined in repositories.json to files_to_keep
+    files_to_keep.extend(self._get_target_repositories())
     # delete files if they no longer correspond to a target defined
     # in targets metadata and are not specified in files_to_keep
-    for root, _, files in os.walk(self.targets_path):
+    for root, _, files in os.walk(str(self.targets_path)):
```
danixeee: I think you can list directory with `self.targets_path.rglob("*")`.

taf/repository_tool.py
```diff
@@ -264,38 +277,41 @@ def add_targets(self, data, targets_role='targets', files_to_keep=None):
...
       for filename in files:
-        filepath = os.path.join(root, filename)
-        if filepath not in data and filename not in files_to_keep:
-          os.remove(filepath)
+        filepath = Path(root) / filename
+        file_rel_path = str(Path(os.path.relpath(str(filepath), str(self.targets_path))).as_posix())
```
danixeee: So, based on previous comment, I think this could simplify things (since you introduced pathlib):

```python
for path in self.targets_path.rglob("*"):
    file_path = path.resolve()
    file_rel_path = str(path)
```

Could you try it out if you have tests?

taf/repository_tool.py
```diff
@@ -264,38 +277,41 @@ def add_targets(self, data, targets_role='targets', files_to_keep=None):
...
 
     for path, target_data in data.items():
       # if the target's parent directory should not be "targets", create
       # its parent directories if they do not exist
-      target_path = os.path.join(self.targets_path, path)
-      if not os.path.exists(os.path.dirname(target_path)):
-        os.makedirs(os.path.dirname(target_path))
+      target_path = self.targets_path / path
+      target_dir = target_path.parents[0]
+      if not target_dir.exists():
```
danixeee: `target_dir.mkdir(parents=True, exist_ok=True)` should work here

taf/repository_tool.py
```diff
@@ -264,38 +277,41 @@ def add_targets(self, data, targets_role='targets', files_to_keep=None):
...
+        os.makedirs(str(target_dir))
 
       # create the target file
       content = target_data.get('target', None)
       if content is None:
-        if not os.path.isfile(target_path):
-          Path(target_path).touch()
+        if not os.path.isfile(str(target_path)):
```
danixeee: `target_path.is_file()`

taf/repository_tool.py
```diff
@@ -264,38 +277,41 @@ def add_targets(self, data, targets_role='targets', files_to_keep=None):
...
+          target_path.touch()
       else:
-        with open(target_path, 'w') as f:
+        with open(str(target_path), 'w') as f:
```
danixeee: `target_path.write_text()`

taf/repository_tool.py
```diff
@@ -306,9 +322,19 @@ def add_targets(self, data, targets_role='targets', files_to_keep=None):
...
+      target_path = self.targets_path / path
+      previous_custom = None
+      if path in previous_targets:
+        previous_custom = previous_targets[path].get('custom')
+      self._add_target(targets_obj, str(target_path), previous_custom)
+
+  def _get_target_repositories(self):
+    repositories_path = self.targets_path / 'repositories.json'
+    if repositories_path.exists():
+      with open(str(repositories_path)) as f:
```
danixeee: `repositories_path.read_text()`

taf/repository_tool.py
renatav: `rglob` seems to work, reworked this loop \

taf/repository_tool.py
renatav: Done

taf/repository_tool.py
renatav: Done

taf/repository_tool.py
renatav: I feel like it's a bit cleaner to keep it like this, because we need to write a json

taf/repository_tool.py
renatav: Changed

PR #45, opened 2019-07-09
Generation of target files, repositories.json fixes and support for ssh urls

### LINKED ISSUE #37: Default namespace value
Default namespace value should be name of the parent directory containing target repositoties

### LINKED ISSUE #39: `add_target_repos` and `generate_repositories_json` should skip auth repo
If the authentication repository is in the same directory as the target directories, it should be skipped. There should be no need to store targets in a separate directory.

### LINKED ISSUE #44: Support ssh urls
This just means that validation of url should not fail in such cases

renatav:
- If the authentication repository is in the same directory as the target directories, it is skipped. This happens both when adding target repos and when generating `repositories.json`
- Using `pathlib` instead of `os`. Support for relative paths
- Setting default `namespace` value. It should be the same as the name of the directory that contains target repositories.
- Validation of repository's urls does not fail if a url is a valid ssh url
- Added tests for validation of urls

Closes #37 
Closes #39 
Closes #44

### DIFF
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -28,7 +28,7 @@ from pathlib import Path
 EXPIRATION_INTERVAL = 36500
 YUBIKEY_EXPIRATION_DATE = datetime.datetime.now() + datetime.timedelta(days=EXPIRATION_INTERVAL)
 
-def add_target_repos(repo_path, targets_directory, namespace=''):
+def add_target_repos(repo_path, targets_directory, namespace=None):
   """
   <Purpose>
     Create or update target files by reading the latest commits of the provided target repositories
@@ -42,6 +42,8 @@ def add_target_repos(repo_path, targets_directory, namespace=''):
   """
   repo_path = Path(repo_path).resolve()
   targets_directory = Path(targets_directory).resolve()
+  if namespace is None:
+    namespace = targets_directory.name
   auth_repo_targets_dir = repo_path / TARGETS_DIRECTORY_NAME
   if namespace:
     auth_repo_targets_dir = auth_repo_targets_dir / namespace
@@ -239,7 +241,7 @@ def generate_keys(keystore, roles_key_infos):
                                        password=password)
 
 
-def generate_repositories_json(repo_path, targets_directory, namespace='',
+def generate_repositories_json(repo_path, targets_directory, namespace=None,
                                targets_relative_dir=None, custom_data=None):
   """
   <Purpose>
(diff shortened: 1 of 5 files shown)


taf/developer_tool.py
```diff
@@ -40,21 +40,25 @@ def add_target_repos(repo_path, targets_directory, namespace=''):
     namespace:
       Namespace used to form the full name of the target repositories. E.g. some_namespace/law-xml
   """
-  auth_repo_targets_dir = os.path.join(repo_path, TARGETS_DIRECTORY_NAME)
+  repo_path = Path(repo_path).resolve()
+  targets_directory = Path(targets_directory).resolve()
+  if namespace is None:
+    namespace = targets_directory.name
+  auth_repo_targets_dir = repo_path / TARGETS_DIRECTORY_NAME
```
danixeee: Can namespace ever be `None` now? If not, you can change lines 47-51. with:

```python
auth_repo_targets_dir = repo_path / TARGETS_DIRECTORY_NAME / namespace
auth_repo_targets_dir.mkdir(parents=True, exist_ok=True)
```

taf/developer_tool.py
```diff
@@ -40,21 +40,25 @@ def add_target_repos(repo_path, targets_directory, namespace=''):
...
-        target_repo_name = os.path.basename(target_repo_dir)
-        with open(os.path.join(auth_repo_targets_dir, target_repo_name), 'w') as f:
-          json.dump({'commit': commit}, f,  indent=4)
+  for target_repo_dir in targets_directory.glob('*'):
+    if not target_repo_dir.is_dir() or target_repo_dir == repo_path:
+      continue
+    target_repo = GitRepository(str(target_repo_dir))
+    if target_repo.is_git_repository:
+      commit = target_repo.head_commit_sha()
+      target_repo_name = os.path.basename(target_repo_dir)
```
danixeee: `target_repo_name  = target_repo_dir.name`

taf/developer_tool.py
```diff
@@ -40,21 +40,25 @@ def add_target_repos(repo_path, targets_directory, namespace=''):
...
+      with open(os.path.join(auth_repo_targets_dir, target_repo_name), 'w') as f:
```
danixeee: `(auth_repo_targets_dir / target_repo_name).write_text(json.dumps({'commit': commit}, indent=4))`, but if you want to keep `with open`, you can just replace `os.path.join` here.

taf/developer_tool.py
```diff
@@ -252,29 +256,35 @@ def generate_repositories_json(repo_path, targets_directory, namespace='',
...
-  auth_repo_targets_dir = os.path.join(repo_path, TARGETS_DIRECTORY_NAME)
-  for target_repo_dir in os.listdir(targets_directory):
-    repo_path = os.path.join(targets_directory, target_repo_dir)
-    if not os.path.isdir(repo_path):
+
+  repo_path = Path(repo_path).resolve()
+  auth_repo_targets_dir = repo_path / TARGETS_DIRECTORY_NAME
+  targets_directory = Path(targets_directory).resolve()
+  if namespace is None:
+    namespace = targets_directory.name
```
danixeee: After this line, `namespace` can't be `None`, so you can remove _if-checks_ in few lines below.

taf/developer_tool.py
```diff
@@ -252,29 +256,35 @@ def generate_repositories_json(repo_path, targets_directory, namespace='',
...
+  for target_repo_dir in targets_directory.glob('*'):
+    if not target_repo_dir.is_dir() or target_repo_dir == repo_path:
       continue
-    target_repo = GitRepository(repo_path)
+    target_repo = GitRepository(target_repo_dir)
```
danixeee: Maybe it will be slightly cleaner if you call `target_repo_dir.resolve()` here, and on line 284. you can replace `str(Path(target_repo.repo_path).resolve())` with `target_repo.repo_path`.

taf/developer_tool.py
renatav: That if is actually intended to check if `namespace` is an empty string. That used to be the default value, making it possible to not specify the namespace at all.

taf/developer_tool.py
renatav: Using `write_text` here seems fine here, will also update `generate_repositories_json`

taf/developer_tool.py
renatav: Same as above, those _if_s are meant to check if `namespace` is an empty string

taf/developer_tool.py
renatav: `str` is necessary in order to support Python 3.5. Will simplify

renatav:
I think that I addressed the comments

PR #48, opened 2019-07-10
Improve Yubikey setup

### LINKED ISSUE #36: Improve YubiKeys setup
- [ ] Ask the user to enter the pin twice
- [ ] If there already is a certificate, ask the user if they want to keep using it or generate a new one
- [ ] Insert the key belonging to [name from cert]

danixeee:
Closes #36.

### DIFF
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -2,31 +2,37 @@ import datetime
 import json
 import os
 import pathlib
-import securesystemslib
-from collections import defaultdict
 from binascii import hexlify
+from collections import defaultdict
+from functools import partial
+from getpass import getpass
+from pathlib import Path
+
+import securesystemslib
+from securesystemslib.exceptions import Error as SSLibError
+from securesystemslib.exceptions import UnknownKeyError
+
 import tuf.repository_tool
+from oll_sc.api import sc_export_x509_pem, sc_sign_rsa_pkcs_pss_sha256
+from oll_sc.exceptions import SmartCardError
+from oll_sc.yk_api import yk_serial_num, yk_setup
+from taf.auth_repo import AuthenticationRepo
+from taf.exceptions import PINMissmatchError
+from taf.git import GitRepository
+from taf.repository_tool import Repository, load_role_key
+from taf.utils import get_pin_for, get_yubikey_pin_for_keyid
+from tuf.keydb import get_key
 from tuf.repository_tool import (METADATA_DIRECTORY_NAME,
                                  TARGETS_DIRECTORY_NAME, create_new_repository,
                                  generate_and_write_rsa_keypair,
+                                 generate_rsa_key,
                                  import_rsa_privatekey_from_file,
                                  import_rsa_publickey_from_file,
-                                 import_rsakey_from_pem,
-                                 generate_rsa_key)
-from tuf.keydb import get_key
-from taf.git import GitRepository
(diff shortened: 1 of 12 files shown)


taf/utils.py
```diff
@@ -33,6 +36,79 @@ def convert(self, value, param, ctx):
...
+  if isinstance(key_slot, int) or isinstance(key_slot, str):
+    key_slot = (int(key_slot),)
+
+  while True:
+    try:
+      input("Please insert {} yubikey and press ENTER.\n".format(holders_name))
+      if not sc_is_present():
+        continue
+
+      key_pin = get_pin_for(holders_name)
```
renatav: No need to confirm the PIN here

taf/developer_tool.py
```diff
@@ -134,74 +140,87 @@ def create_repository(repo_path, keystore, roles_key_infos, commit_message=None)
...
+                      key_id_certs[key['keyid']] = cert_cn
+                      print('{} id: {}'.format(cert_cn, key['keyid']))
+                  existing_keyid = input("Enter existing Yubikey's id and press ENTER ")
+                  try:
+                    key = get_key(existing_keyid)
+                    cert_cn = key_id_certs[existing_keyid]
+                  except UnknownKeyError:
+                    pass
+            if not use_existing:
+              input("Please insert a new Yubikey and press ENTER.")
```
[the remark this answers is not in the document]
danixeee: When pins do not match, user will be asked to insert them again now.

taf/developer_tool.py
```diff
@@ -136,41 +140,43 @@ def create_repository(repo_path, keystore, roles_key_infos, commit_message=None)
...
                 pass
         if not use_existing:
-          input("Please insert a new YubiKey and press ENTER.")
+          input("Please insert a new Yubikey and press ENTER.")
           serial_num = yk_serial_num()
           while serial_num in yubikeys[role_name]:
-            print("YubikKy with serial number {} is already in use.\n".format(serial_num))
-            input("Please insert new YubiKey and press ENTER.")
+            print("Yubikey with serial number {} is already in use.\n".format(serial_num))
+            input("Please insert new Yubikey and press ENTER.")
```
renatav: a new Yubikey

taf/utils.py
```diff
@@ -33,6 +36,84 @@ def convert(self, value, param, ctx):
...
+  while True:
+    try:
+      input("Please insert {}'s yubikey and press ENTER.\n".format(holders_name))
+      if not sc_is_present():
+        continue
+
+      key_pin = get_pin_for(holders_name, confirm=False)
+      inserted_key = get_yubikey_public_key(key_slot, key_pin)
+      if inserted_key['keyid'] not in expected_keyids:
+        print("Please insert valid yubikey!")
```
renatav: When setting up repository, if a user's inserts root1 Yubikey instead of root2 yubikey when prompted to do so, he will be asked to enter a pin and then asked to insert a valid Yubikey.
If that's not hard to do, can we check if the key matches the expected one before a user is asked to insert the PIN? Also, valid might not be the best word here, since the inserted Yubikey is valid, but not the one we expect

danixeee: It is possible, but I would need to change `oll-sc`. Since we want to get rid of `oll-sc` and move existing yubikey functions to TAF, I will address this comment after that work.

PR #67, opened 2019-09-23
Updated default logs location

### LINKED ISSUE #64: Configure location of taf.log and taf.err

renatav:
TAF  logs used to be created wherever taf was called. This resulted in these logs occurring everywhere and had to be added to gitignore. It is now possible to specify their location. By default, TAF logs
will be placed inside `~/.taf` on Windows and `/var/log` on Linux/macOS

Updated travis build script since permission denied errors occurred when running tests.

Closes #64

### DIFF
--- a/taf/log.py
+++ b/taf/log.py
@@ -11,16 +11,14 @@ formatter = logging.Formatter(_FORMAT_STRING)
 logger = logging.getLogger('taf')
 logger.setLevel(taf.settings.LOG_LEVEL)
 
+
 def _get_log_location():
-  location = taf.settings.LOGS_LOCATION
+  location = taf.settings.LOGS_LOCATION or os.environ.get('TAF_LOG')
   if location is None:
-    if os.name == 'nt':
-      location = Path.home() / '.taf'
-      location.mkdir(exist_ok=True)
-    else:
-      location = Path('/var/log')
+    location = Path.home() / '.taf'
+    location.mkdir(exist_ok=True)
   else:
-    location = Path(logs_location)
+    location = Path(location)
   return location
 
 
--- a/taf/log.py
+++ b/taf/log.py
@@ -1,4 +1,5 @@
 import logging
+from pathlib import Path
 
 import taf.settings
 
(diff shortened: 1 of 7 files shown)


taf/log.py
```diff
@@ -9,23 +10,34 @@
...
   file_handler.setLevel(taf.settings.FILE_LOGGING_LEVEL)
   file_handler.setFormatter(formatter)
   logger.addHandler(file_handler)
 
-if taf.settings.SEPARATE_ERRORS:
-  error_handler = logging.FileHandler(taf.settings.ERROR_LOF_FILENAME)
-  error_handler.setLevel(taf.settings.ERROR_LOGGING_LEVEL)
-  error_handler.setFormatter(formatter)
-  logger.addHandler(error_handler)
+  if taf.settings.SEPARATE_ERRORS:
```
danixeee: Should lines 36-40. be indented?

taf/log.py
```diff
@@ -9,23 +10,34 @@
...
   console_handler = logging.StreamHandler()
   console_handler.setLevel(taf.settings.CONSOLE_LOGGING_LEVEL)
   console_handler.setFormatter(formatter)
   logger.addHandler(console_handler)
 
+
 if taf.settings.ENABLE_FILE_LOGGING:
-  file_handler = logging.FileHandler(taf.settings.LOG_FILENAME)
+  logs_location = taf.settings.LOGS_LOCATION
+  if logs_location is None:
```
danixeee: Could you extract this to a function (`_get_log_location`)? Maybe it would be better to check if platform is windows and use `Path.home() / '.taf'` in that case.

taf/log.py
renatav: I moved this `if  taf.settings.SEPARATE_ERRORS` inside `if taf.settings.ENABLE_FILE_LOGGING` (doesn't really make sense to create `.err` if we should not log to files.

danixeee: I see now, never-mind.

taf/log.py
```diff
@@ -9,23 +11,38 @@
 logger = logging.getLogger('taf')
 logger.setLevel(taf.settings.LOG_LEVEL)
 
+def _get_log_location():
+  lcation = taf.settings.LOGS_LOCATION
```
danixeee: lcation -> location

PR #70, opened 2019-10-09
General refactoring

### LINKED ISSUE #54: Repository update improvements
- `add_target_repos` should be able to traverse to `repositories.json` entries and add targets based on that
- rename `add_targets` to `sign_targets`
- rename `add_target_repos` to something like `update_targets`
- Check if everything works when when loading snapshot and timestamp keys from the file system and using YubiKey for signing

### LINKED ISSUE #59: Consistently use pathlib instead of the os module
The majority of `developer_tool` relies on `pathlib`. The exceptions are `build_repository` and `create_repository` functions. Additionally, `os.makedirs` can be replaced with Path's `mkdir`.  We can also use glob with `**`, which should return all this directory and all subdirectories recursively. Since we are transitioning to glob, it might be better to use in places where we later check if a path is a directory. We should also update other modules, like `repository_tool`, and the updater.

danixeee:

Closes #54.
Closes #59.

+ some more refactoring.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
 ignore = E203, E266, E501, W503
 max-line-length = 88
-max-complexity = 20
+max-complexity = 18
 select = B,C,E,F,W,T4,B9
--- a/taf/cli.py
+++ b/taf/cli.py
@@ -22,7 +22,7 @@ def cli():
 @click.option('--commit-msg', default=None, help='Commit message to be used in case the changes'
               'should be automatically committed')
 @click.option('--scheme', default=DEFAULT_RSA_SIGNATURE_SCHEME, help='A signature scheme used for signing.')
-def add_targets(repo_path, keystore, keys_description, commit_msg, scheme):
+def sign_targets(repo_path, keystore, keys_description, commit_msg, scheme):
     developer_tool.register_target_files(repo_path, keystore, keys_description, commit_msg, scheme)
 
 
--- a/taf/cli.py
+++ b/taf/cli.py
@@ -49,6 +49,9 @@ def update_repos_from_fs(repo_path, targets_dir, namespace):
 
 @cli.command()
 @click.option("--repo-path", default="repository", help="Location of the repository")
+@click.option('--targets-dir', default='targets', help='Directory where the target '
+              'repositories are located')
+@click.option('--namespace', default=None, help='Namespace of the target repositories')
 def update_repos_from_repositories_json(repo_path, targets_dir, namespace):
     "Updates target repositories by traversing repositories.json "
     developer_tool.update_target_repos_from_repositories_json(repo_path)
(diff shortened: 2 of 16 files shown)


taf/cli.py
```diff
@@ -42,8 +42,16 @@ def add_target_file(repo_path, file_path, keystore, keys_description, scheme):
...
 @click.option('--namespace', default=None, help='Namespace of the target repositories')
-def add_target_repos(repo_path, targets_dir, namespace):
-    developer_tool.add_target_repos(repo_path, targets_dir, namespace)
+def update_repos_from_fs(repo_path, targets_dir, namespace):
+    "Updates target repositories by traversing given targets directory "
+    developer_tool.update_target_repos_from_fs(repo_path, targets_dir, namespace)
+
+
+@cli.command()
+@click.option("--repo-path", default="repository", help="Location of the repository")
```
renatav: Missing `targets_dir` and `namespace` click options

taf/developer_tool.py
```diff
@@ -134,27 +154,107 @@ def build_auth_repo(
...
+    # if it does not, generate the keys and print the output
+    else:
+        key = generate_rsa_key()
+        print(f"{role_name} key:\n\n{key['keyval']['private']}\n\n")
+        public_key = private_key = key
+    role_obj.add_verification_key(public_key)
+    role_obj.load_signing_key(private_key)
+
+
 def create_repository(
```
renatav: Definitely nothing crucial, but we there is one `os.path.isdir` left in this function

taf/developer_tool.py
```diff
@@ -134,27 +154,107 @@ def build_auth_repo(
...
+
+    # set Yubikey expiration date
+    role_obj.add_verification_key(key, expires=YUBIKEY_EXPIRATION_DATE)
+    role_obj.add_external_signature_provider(
+        key, partial(signature_provider, key["keyid"], cert_cn)
+    )
+    yubikeys[role_name][serial_num] = (key, cert_cn)
+
+
+def _register_key(keystore, passwords, role_obj, role_name, key_name, key_num, scheme):
```
renatav: We could also replace these few `os.path` calls here.

taf/developer_tool.py
```diff
@@ -281,22 +302,22 @@ def create_repository(
 
 def generate_keys(keystore, roles_key_infos):
```
renatav: One `os.path` call left here

.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
 ignore = E203, E266, E501, W503
 max-line-length = 88
-max-complexity = 20
+max-complexity = 18
```
renatav: Did you mean to commit this? Is is better t reduce complexity?

danixeee: That's the number of nested statements, lower is better for code readability.

PR #79, opened 2019-11-21
Creation and setup of taf repositories enhancements

renatav:

- Added new commands for:
  - setting up a new yubikey
  - exporting a public key from inserted yubikey
  - adding a new signing key (currently assumes that all keys are stored on yubikeys)

- Improved creation of repositories:
  - check keystore before inserting all of yubikeys
  - created standardized yubikey prompt
  -  keep retrying to access the yubikey rather than failing out with an exception (can take a few seconds for computer to recognize yubikey after inserted)
  -  store pins globallyso don't have to keep typing password.
  - made the process more robust
  - added support for mixing keys stored on yubikeys with keys from keystore files

### DIFF
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -8,6 +8,7 @@ from getpass import getpass
 from pathlib import Path
 
 import securesystemslib
+import click
 import tuf.repository_tool
 from securesystemslib.exceptions import UnknownKeyError
 from securesystemslib.interface import (
@@ -56,24 +57,36 @@ def add_signing_key(repo_path, role, pub_key_path):
     threshold = root_obj.threshold
     num_of_signatures = 0
 
-    while True:
+    def _read_role_key(role):
         input(f"Please insert {role} YubiKey and press enter")
         role_public_key = yk.get_piv_public_key_tuf()
         if not taf_repo.is_valid_metadata_yubikey(role, role_public_key):
             print(f"The inserted YubiKey is not a valid {role} key")
-        else:
-            taf_repo.add_yubikey_external_signature_provider(role, role)
-            break
+            return False
+        taf_repo.add_yubikey_external_signature_provider(role, role)
+        return True
 
-    while num_of_signatures < threshold:
-        name = f"root{num_of_signatures+1}"
-        input(f"Please insert {name} YubiKey and press enter")
-        root_public_key = yk.get_piv_public_key_tuf()
-        if not taf_repo.is_valid_metadata_yubikey("root", root_public_key):
-            print("The inserted YubiKey is not a valid root key")
-            continue
-        taf_repo.add_yubikey_external_signature_provider("root", name)
-        num_of_signatures += 1
+    while not _read_role_key(role):
+        pass
+
+    all_loaded = False
+    while not all_loaded:
+        if num_of_signatures >= threshold:
+            all_loaded = not (
+                click.confirm(
+                    "Threshold of root keys reached. Do you want to load more root keys?"
+                )
+            )
+        # TODO check if key already loaded
+        if not all_loaded:
+            name = f"root{num_of_signatures+1}"
+            input(f"Please insert {name} YubiKey and press enter")
+            root_public_key = yk.get_piv_public_key_tuf()
+            if not taf_repo.is_valid_metadata_yubikey("root", root_public_key):
+                print("The inserted YubiKey is not a valid root key")
+                continue
+            taf_repo.add_yubikey_external_signature_provider("root", name)
+            num_of_signatures += 1
 
     taf_repo.writeall()
 
@@ -189,16 +202,17 @@ def build_auth_repo(
 
 def _register_yubikey(yubikeys, role_obj, role_name, key_name, scheme, certs_dir):
     from taf.repository_tool import yubikey_signature_provider
+
     print(f"Registering keys for {key_name}")
     use_existing = input("Do you want to reuse already set up Yubikey? y/n ") == "y"
     if use_existing:
-            input("Please insert an already set up YubiKey and press ENTER")
-            key = yk.get_piv_public_key_tuf()
-            cert_cn = input("Enter key holder's name: ")
-            serial_num = yk.get_serial_num()
+        input("Please insert an already set up YubiKey and press ENTER")
+        key = yk.get_piv_public_key_tuf()
+        cert_cn = input("Enter key holder's name: ")
+        serial_num = yk.get_serial_num()
 
     else:
-        print('WARNING - this will delete everything from the inserted key')
+        print("WARNING - this will delete everything from the inserted key")
         input("Please insert a new YubiKey and press ENTER.")
         serial_num = yk.get_serial_num()
         while serial_num in yubikeys[role_name]:
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -47,6 +47,7 @@ YUBIKEY_EXPIRATION_DATE = datetime.datetime.now() + datetime.timedelta(
 
 def add_signing_key(repo_path, role, pub_key_path):
     from taf.repository_tool import yubikey_signature_provider
+
     taf_repo = Repository(repo_path)
     pub_key_pem = Path(pub_key_path).read_text()
     taf_repo.add_metadata_key(role, pub_key_pem, DEFAULT_RSA_SIGNATURE_SCHEME)
@@ -71,7 +72,7 @@ def add_signing_key(repo_path, role, pub_key_path):
         if not taf_repo.is_valid_metadata_yubikey("root", root_public_key):
             print("The inserted YubiKey is not a valid root key")
             continue
-        taf_repo.add_yubikey_external_signature_provider('root', name)
+        taf_repo.add_yubikey_external_signature_provider("root", name)
         num_of_signatures += 1
 
     taf_repo.writeall()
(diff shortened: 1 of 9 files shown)


taf/yubikey.py
```diff
@@ -19,12 +20,25 @@
...
 
 DEFAULT_PIN = "123456"
 DEFAULT_PUK = "12345678"
 
 
+_pins_dict = {}
+
+
+def add_key_pin(serial_num, pin):
+    global _pins_dict
```
danixeee: No need for `global` here.

taf/yubikey.py
```diff
@@ -296,3 +310,101 @@ def setup(
...
+            role,
+            taf_repo,
+            registering_new_key,
+            creating_new_key,
+            loaded_yubikeys,
+            pin_confirm,
+            pin_repeat,
+            prompt_message,
+        )
+        if success:
```
danixeee: Should we raise exception if `success` is `False` or if not, we should `return None, None`?

taf/repository_tool.py
```diff
@@ -335,6 +370,14 @@ def add_targets(self, data, targets_role="targets", files_to_keep=None):
             if target_path.is_file():
                 self._add_target(targets_obj, str(target_path), previous_custom)
 
+    def add_yubikey_external_signature_provider(self, role, name):
```
danixeee: Not used anymore

taf/exceptions.py
```diff
@@ -39,6 +43,10 @@ def __init__(self, message):
         super().__init__("root", message)
 
 
+class KeystoreError(TAFError):
```
danixeee: Does it make sense to subclass `YubikeyError`?

taf/yubikey.py
```diff
@@ -296,3 +310,101 @@ def setup(
...
+def get_and_validate_pin(key_name, pin_confirm=True, pin_repeat=True):
+    valid_pin = False
+    while not valid_pin:
+        pin = get_pin_for(key_name, pin_confirm, pin_repeat)
+        valid_pin, retries = is_valid_pin(pin)
+        if not valid_pin and not retries:
+            raise InvalidPINError("No retries left. YubiKey locked.")
+        if not valid_pin:
+            if not click.confirm(
+                f"Incorrect PIN. Do you want to try again? {retries} retires left."
```
danixeee: `retires` -> `retries`

taf/developer_tool.py
```diff
@@ -45,6 +45,127 @@
...
+                    click.confirm(
+                        f"Threshold of {role} keys reached. Do you want to load more root keys?"
+                    )
+                )
+            if not all_loaded:
+                is_yubikey = None
+                if role_key_infos is not None and role in role_key_infos:
+                    is_yubikey = role_key_infos[role].get("yubikey")
+                if is_yubikey is None:
+                    is_yubikey = click.confirm(f"Load {key_name} from YubiKey?")
```
danixeee: I got confused when I saw this question. Should we ask something like: "Do you want to sing {key_name} with Yubikey?" ?

taf/developer_tool.py
```diff
@@ -45,6 +45,127 @@
...
+                if is_yubikey:
+                    yk.yubikey_prompt(key_name, role, taf_repo, False, loaded_yubikeys)
+                else:
+                    pem = getpass(
+                        f"Enter {key_name} private key without its header and footer"
```
danixeee: Could you add a new line to the end of a message?

taf/developer_tool.py
```diff
@@ -45,6 +45,127 @@
...
+                    )
+                    pem = _form_private_pem(pem)
```
danixeee: Does it make sense to loop until a valid PEM is entered? If I enter something like "abcde" I get following error `securesystemslib.exceptions.CryptoError: RSA (public, private) tuple cannot be generated from the encrypted PEM string: Could not deserialize key data.`

taf/developer_tool.py
danixeee: `loaded_yubikeys` is passed as a wrong argument. `yk.yubikey_prompt(key_name, role, taf_repo, loaded_yubikeys=loaded_yubikeys)`

taf/developer_tool.py
```diff
@@ -155,76 +276,52 @@ def build_auth_repo(
...
+    keystore, roles_key_info, role_obj, role_name, key_name, key_num, scheme
+):
     # if keystore exists, load the keys
     # this is useful when generating tests
     if keystore is not None:
-        public_key = import_rsa_publickey_from_file(
-            str(Path(keystore, f"{key_name}.pub")), scheme
+        public_key = _read_public_key_from_keystore(keystore, key_name)
+        private_key = _read_private_key_from_keystore(
+            keystore, key_name, roles_key_info, key_num
```
danixeee: `scheme` should be passed when loading a key

taf/developer_tool.py
```diff
@@ -155,76 +276,52 @@ def build_auth_repo(
...
-                str(Path(keystore, key_name)), password, scheme=scheme
-            )
-        else:
-            private_key = import_rsa_privatekey_from_file(
-                str(Path(keystore, key_name)), scheme=scheme
-            )
-
     # if it does not, generate the keys and print the output
     else:
         key = generate_rsa_key()
```
danixeee: `scheme` should be passed when generating a key

taf/developer_tool.py
```diff
@@ -547,6 +733,37 @@ def _check_key_id(expected_key_id):
...
+        "new Yubikey",
+        creating_new_key=True,
+        pin_confirm=True,
+        pin_repeat=True,
+        prompt_message="Please insert the new Yubikey and press ENTER",
+    )
+    _setup_new_yubikey(serial_num, certs_dir)
+
+
+def _setup_new_yubikey(serial_num, certs_dir=None, scheme=DEFAULT_RSA_SIGNATURE_SCHEME):
```
danixeee: `key` should be returned from this function

taf/developer_tool.py
```diff
@@ -155,76 +276,52 @@ def build_auth_repo(
...
-    # set Yubikey expiration date
-    role_obj.add_verification_key(key, expires=YUBIKEY_EXPIRATION_DATE)
-    role_obj.add_external_signature_provider(
-        key, partial(signature_provider, key["keyid"], cert_cn)
-    )
-    yubikeys[role_name][serial_num] = (key, cert_cn)
+        registered = True
+        # set Yubikey expiration date
+        # add_yubikey_external_signature_provider
+        role_obj.add_verification_key(key, expires=YUBIKEY_EXPIRATION_DATE)
```
danixeee: `key` is `None` if I choose to setup a new yubikey

taf/developer_tool.py
```diff
@@ -547,6 +733,37 @@ def _check_key_id(expected_key_id):
...
+
+
+def _setup_new_yubikey(serial_num, certs_dir=None, scheme=DEFAULT_RSA_SIGNATURE_SCHEME):
+    pin = yk.get_key_pin(serial_num)
+    cert_cn = input("Enter key holder's name: ")
+    print("Generating key, please wait...")
+    pub_key_pem = yk.setup(pin, cert_cn, cert_exp_days=EXPIRATION_INTERVAL).decode(
+        "utf-8"
+    )
+    scheme = DEFAULT_RSA_SIGNATURE_SCHEME
```
danixeee: This line should be removed

taf/developer_tool.py
```diff
@@ -547,6 +733,37 @@ def _check_key_id(expected_key_id):
...
+    key = import_rsakey_from_pem(pub_key_pem, scheme)
+    if certs_dir is None:
```
danixeee: Could we move lines 758-766 to a new function, so we can export certificate to `certs_dir` when using existing yubikeys?

taf/yubikey.py
renatav: The idea is to repeat `_read_and_check_yubikey` until success is `True`.

taf/exceptions.py
renatav: That seems confusing to me, but it would make sense to create a common ancestor class. Something like `KeyError`, sublassed by `KeystoreError` and `YubikeyError`?

danixeee: I guess we can live it like this then

renatav: If at some point we need to catch both exceptions, we should consider creating a superclass. At the moment, that does not occur anywhere

taf/developer_tool.py
renatav: Moved all this code to a new function where we loop until a valid PEM is entered and check if it's a valid key of the given role

taf/developer_tool.py
renatav: There was a return missing in `setup_new_yubikey`.

taf/developer_tool.py
renatav: Done

renatav:
@danixeee I think that I addressed all comments. Tested all variations - new yubikeys, existing yubikeys, loading keys from keystore and entering keys

danixeee:
@renatav When I follow the same flow for testing (https://github.com/openlawlibrary/taf/pull/79#pullrequestreview-321621916) I cannot sign targets with the following command `taf sign_targets --repo-path ./law --keystore ./data/keystore/` and the reason is that _targets_ key is on my yubikey and not in a keystore (and cannot be exported). I think we should slightly update the function to use keystore for keys that are not generated on yubikeys (that is specified in _keys.json_) and not all keys. Does this make sense or `--keystore` flag means that _all_ keys will be loaded from files?
Other than that, everything seems fine and if this comment does not make sense, we can merge this.

PR #97, opened 2020-02-01
Support delegated targets roles

### LINKED ISSUE #52: Support targets delegations
At the moment, we assume that we have only one targets role (the root targets role `targets`). Update all places where that is the case and support delegated roles. We also need to support creation of delegated roles.  The users should be able to specify which targets role is responsible for which paths. We also need to be able to set up YubiKeys for these roles.

renatav:

Added support for delegated targets roles

- Support creation of delegations when creating a new repository
- Implemented updated of metadata files corresponding to delegated roles
- Refactored and updated metadata signing - each role can be signed with multiple keys, which can be loaded from disk or kept on Yubikeys

Closes #52

### DIFF
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -306,8 +306,6 @@ def create_repository(
 
     # load and/or generate all keys first
     try:
-        # load keys not stored on YubiKeys first, to avoid entering pins
-        # if there is somethig wrong with keystore files
         keystore_roles, yubikey_roles = _sort_roles(roles_key_infos, repository)
         signing_keys = {}
         verification_keys = {}
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -295,9 +295,10 @@ def create_repository(
                 keystore_roles.append((role_name, role_key_info))
             else:
                 yubikey_roles.append((role_name, role_key_info))
-            if 'delegations' in role_key_info:
-                delegated_keystore_role, delegated_yubikey_roles = \
-                    _sort_roles(role_key_info['delegations'], repository)
+            if "delegations" in role_key_info:
+                delegated_keystore_role, delegated_yubikey_roles = _sort_roles(
+                    role_key_info["delegations"], repository
+                )
                 keystore_roles.extend(delegated_keystore_role)
                 yubikey_roles.extend(delegated_yubikey_roles)
         return keystore_roles, yubikey_roles
@@ -310,14 +311,21 @@ def create_repository(
         signing_keys = {}
         verification_keys = {}
         for role_name, key_info in keystore_roles:
-            keystore_keys, _ = _setup_roles_keys(role_name, key_info, repository, keystore=keystore)
+            keystore_keys, _ = _setup_roles_keys(
+                role_name, key_info, repository, keystore=keystore
+            )
             for public_key, private_key in keystore_keys:
                 signing_keys.setdefault(role_name, []).append(private_key)
                 verification_keys.setdefault(role_name, []).append(public_key)
 
         for role_name, key_info in yubikey_roles:
-            _, yubikey_keys = _setup_roles_keys(role_name, key_info, repository, certs_dir=auth_repo.certs_dir,
-                                                yubikeys=yubikeys)
+            _, yubikey_keys = _setup_roles_keys(
+                role_name,
+                key_info,
+                repository,
+                certs_dir=auth_repo.certs_dir,
+                yubikeys=yubikeys,
+            )
             verification_keys[role_name] = yubikey_keys
 
     except KeystoreError as e:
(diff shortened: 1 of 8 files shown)


taf/tests/test_repository_tool.py
```diff
@@ -144,10 +120,11 @@ def test_update_targets_valid_key_valid_pin(taf_happy_path, targets_yk):
         "dummy/target_dummy_repo": {"target": {"commit": target_commit_sha}},
         "capstone": {},
     }
-
+    yk.add_key_pin(targets_yk.serial, "123456")
```
danixeee: Could you please replace all `"123456"` strings with `yubikey_utils.VALID_PIN` contant?

taf/repository_tool.py
```diff
@@ -395,7 +405,7 @@ def _get_target_repositories(self):
             repositories = json.loads(repositories)["repositories"]
             return [str(Path(target_path).as_posix()) for target_path in repositories]
 
-    def get_role_keys(self, role):
+    def get_role_keys(self, role, parent_role=None):
         """Registers new target files with TUF.
```
danixeee: The old docstring is wrong :)

taf/repository_tool.py
```diff
@@ -411,10 +421,18 @@ def get_role_keys(self, role):
                                                         targets object.
         """
         role_obj = self._role_obj(role)
-        return role_obj.keys
+        try:
+            return role_obj.keys
+        except KeyError:
```
danixeee: If `role_obj` is `None`, we should catch `AttributeError`

taf/repository_tool.py
```diff
@@ -593,188 +625,435 @@ def update_root(self, signature_dict):
         except (TUFError, SSLibError) as e:
             raise RootMetadataUpdateError(str(e))
 
-    def update_snapshot(
+    def sign_role_keystores(self, role_name, signing_keys, write=True):
```
danixeee: `sign_role_keystore` sound better to me. Parameter name `signing_keys` indicates that there could be multiple keys used.

taf/repository_tool.py
```diff
@@ -593,188 +625,435 @@ def update_root(self, signature_dict):
...
+            raise SigningError(
+                role_name,
+                f"Insufficient number of signing keys. Signing threshold is {threshold}.",
+            )
+        for key in signing_keys:
+            self._try_load_metadata_key(role_name, key)
+        if write:
+            self._repository.write(role_name)
+
+    def sign_role_yubikeys(
```
danixeee: Same as for previous method. What do you think?

taf/repository_tool.py
renatav: Yep, will fix it

taf/repository_tool.py
renatav: This does support multiple keys. `signing_keys` is expected to be a list of keys (see how this is called from the `developer_tool`). I was thinking that we can readd methods which would expect just one key/one pin, so that we don't have to call this with a list with only a single contained item. Would that be useful?

danixeee: I guess it's fine to call with a list that contains only one key? But I wanted to say that all keys are from the same keystore (are they?), so the name of this and other methods' names that ends with "keystores" and "yubikeys" should be "keystore" and "yubikey". Does it make sense?

renatav: `_keystores` does not sound that great to me, I agree. The keys can also be entered through the command line, so we don't even have to have keystore files. But they are loaded from the file system, they are loaded from different files which are all expected to be inside the same directory. I don't know what would be the best name. At one point I named these methods `_keys` and `_yubikeys` but that also sounds confusing. `_yubikeys` sounds good to me because we can use different yubikeys. But `_keystore` and `_yubikeys` also sounds confusing.

danixeee: Hm, I am not sure then, we can keep it like this until we figure out better naming.

taf/developer_tool.py
```diff
@@ -345,115 +277,290 @@ def create_repository(
...
         # load keys not stored on YubiKeys first, to avoid entering pins
-        # if there is something wrong with keystore files
-        for role_name, key_info in roles_key_infos.items():
-            if not key_info.get("yubikey", False):
-                _register_roles_keys(role_name, key_info, repository)
-
-        for role_name, key_info in roles_key_infos.items():
-            if key_info.get("yubikey", False):
-                _register_roles_keys(role_name, key_info, repository)
+        # if there is somethig wrong with keystore files
```
danixeee: same comment as on [line 290](https://github.com/openlawlibrary/taf/pull/97/files#diff-a0a191c70a0f3c2446286f9c322e5404R290).

taf/repository_tool.py
```diff
@@ -676,13 +680,20 @@ def sign_role_yubikeys(
...
                 f"Insufficient number of signing keys. Signing threshold is {threshold}.",
             )
+
+        if pins is not None:
+            for serial_num, pin in pins.items():
+                yk.add_key_pin(serial_num, pin)
+
         for index, public_key in enumerate(public_keys):
             if public_key is None:
                 from taf.yubikey import get_piv_public_key_tuf
```
danixeee: This import statement should be moved outside the for loop, perhaps to line 684?

taf/repository_tool.py
danixeee: Did you forget to change this?

taf/repository_tool.py
```diff
@@ -411,10 +423,38 @@ def get_role_keys(self, role):
...
+        Raises:
+        - securesystemslib.exceptions.FormatError: If the arguments are improperly formatted.
+        - securesystemslib.exceptions.UnknownRoleError: If 'rolename' has not been delegated by this
+        """
+        role_obj = self._role_obj(role)
+        if role_obj is None:
+            return None
+        try:
+            return role_obj.threshold
+        except KeyError:
```
danixeee: `AttributeError` here as well

PR #101, opened 2020-02-25
Extended updater to support delegations and further delegations fixes

### LINKED ISSUE #99: Updater - add support for delegated roles when validating targets
Delegations should not impact update of validation repositories, but target repositories are validated based on `targets.json` which should no longer be the case

renatav:

- Extended the updater so that it can validate target repositories in cases when delegated roles are responsible for signing the corresponding targets files. 
- In order to avoid copy-pasting code from `repository_tool`'s `Repository` class, `AuthRepo` now also inherits the mentioned class. This makes sense because authentication repositories are both git and TUF repositories (their structure corresponds to that required by TUF). This will also allow simplification of code in places where we had to instantiate `Repository` in addition to already having an instance of the `AuthRepo` class
- Added new updater tests - one where authentication repository with delegated targets roles and its targets repositories can be successfully validated and once when update is supposed to fail.
-Fixed an issue with registering targets in case when we only want to update targets of one delegated role. TUF would automatically register other target files and calculate lengths and shas of these files without removing the new line at the end of file.

To test the updater call:
` taf repo update https://github.com/openlawlibrary/smc-law-test location\smc-law-test --authenticate-test-repo`

Closes #99


taf/repository_tool.py
```diff
@@ -698,11 +780,12 @@ def map_signing_roles(self, target_filenames):
         is expected to be relative to the targets directory. It can be defined as a glob
         pattern.
         """
+        if delegations_info_fn is None:
```
danixeee: Could you add docstring for `delegations_info_fn`? Or no needed if we could implement a method to return a repository at some other state (commit sha).

taf/repository.py
```diff
@@ -0,0 +1,4 @@
+class BaseRepository:
```
danixeee: Should we move this to `repository_tool.py` module?

taf/auth_repo.py
```diff
@@ -164,35 +181,26 @@ def _safely_get_json(self, commit, path):
...
-        targets_path="targets",
         repo_urls=None,
         additional_info=None,
         default_branch="master",
     ):
-
         super().__init__(
-            root_dir, repo_name, repo_urls, additional_info, default_branch
+            root_dir=root_dir,
+            repo_name=repo_name,
```
danixeee: Looks good overall. Two main things that wee should consider:

- standardize parameter names - e.g. `repo_name`, `repository_name`, `repo`, `repository` (if class is called `Repository`, just `name` should be enough, but in case of multi-inheritance we could name it `repo_name` in subclass)
- creating a mechanism to create a `Repository` instance in previous revisions (contextmanager, pathing a tuf util function which reads metadata files)

taf/repository_tool.py
```diff
@@ -42,6 +43,14 @@
 DISABLE_KEYS_CACHING = False
 
 
+def get_role_metadata_path(role):
```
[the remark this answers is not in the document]
renatav: The difference between `targets_path ` and `metadata_path` and these two functions is that the first pair return full path (`repo_path` + metadata/targets dir), while the second pair is independent of `repo_path` - they return paths of files relative to the repository's root. We need this when loading files using git. I think that we should keep the two properties in any case. I moved `get_role_metadata_path` and `get_target_path` so that we wouldn't have to import `METADATA_DIRECTORY_NAME` and `TARGETS_DIRECTORY_NAME` in other modules as well. We could turn them into static methods, but I've been avoiding that since reading some posts about that being a bad practice in Python

taf/repository_tool.py
```diff
@@ -134,31 +143,36 @@ def _check_key_and_get_pin(expected_key_id):
...
+    @property
+    def _repository(self):
+        if self._tuf_repository is None:
+            self._tuf_repository = load_repository(self.repo_path, self.repo_name)
+        return self._tuf_repository
 
     @property
     def repo_id(self):
-        return GitRepository(self.repository_path).initial_commit
+        return GitRepository(self.repo_path).initial_commit
```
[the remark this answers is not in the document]
renatav: I think that we only use this in the management application, so I am not sure if a combination with `GitRepository` is needed there. But I agree, this does make sense

taf/repository_tool.py
```diff
@@ -592,6 +662,11 @@ def get_role_repositories(self, role, parent_role=None):
             if all([fnmatch(repo, path) for path in role_paths])
         ]
 
+    def get_delefations_info(self, role_name):
+        # load repository is not already loaded
+        self._repository
```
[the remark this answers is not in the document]
renatav: The problem is that that would prevent us from instantiating an empty `AuthRepo` because `tuf` raises an error if mandatory metadata files do not exist yet.

taf/repository.py
renatav: I thought about where to put this class a lot. Still not sure. The idea behind `repository_tool` is to link our TAF Repository with TUF repository. So that does not seem like the best place. I initially thought that there would be more code in this module. Maybe it makes sense to move it to `repository_tool` and move it elsewhere if there is a need to extend it.

taf/repository_tool.py
renatav: I think that it makes sense to have such a functionality, unless it adds to much complexity to our code base since we must avoid modifying TUF. I'll try to implement something, but will timebox it to say 2-3 hours since we don't really need this functionality in other places at the moment. If I don't make much progress, I'll create an issue and we can address it when there is a greater need.

taf/repository.py
danixeee: Or we could remove it completely for now since it does not have any logic.

renatav: Reoved

taf/auth_repo.py
renatav: I believe that all repository classes contain the word `Repository`. I renamed `repo_name` to `name` and `repo_path` to `path`. We'll need to update all places in our code that extend `GitRepository`. Added a mechanism to instantiate TUF repository given an older commit

taf/git.py
```diff
@@ -26,22 +31,24 @@ def __init__(
       additional_info: a dictionary containing other data (optional)
       default_branch: repository's default branch
     """
-        self.repo_path = str(Path(repo_path).resolve())
+        self.path = str(Path(repo_path).resolve())
```
danixeee: What do you think about changing this to `self._path = Path(repo_path).resolve()` and adding a property called `path`? There are some places in code where we could use `_path` as _pathlib.Path_ instance (comments below). That way we will keep backwards compatibility.

taf/git.py
```diff
@@ -26,22 +31,24 @@ def __init__(
...
+        if repo_name is None:
+            repo_name = os.path.basename(self.path)
```
danixeee: `repo_name = self._path.name`

taf/git.py
```diff
@@ -54,7 +61,7 @@ def remotes(self):
     @property
     def is_git_repository_root(self):
         """Check if path is git repository."""
-        git_path = Path(self.repo_path) / ".git"
+        git_path = Path(self.path) / ".git"
```
danixeee: `self._path / '.git'`

taf/git.py
```diff
@@ -283,9 +286,9 @@ def clean(self):
 
     def clone(self, no_checkout=False, bare=False):
 
-        taf_logger.info("Repo {}: cloning repository", self.repo_name)
-        shutil.rmtree(self.repo_path, True)
-        os.makedirs(self.repo_path, exist_ok=True)
+        taf_logger.info("Repo {}: cloning repository", self.name)
+        shutil.rmtree(self.path, True)
+        os.makedirs(self.path, exist_ok=True)
```
danixeee: `self._path.mkdir(exists=True, parents=True)` (rmdir for above line?)

taf/git.py
```diff
@@ -475,8 +465,8 @@ def get_tracking_branch(self, branch=""):
             return None
 
     def init_repo(self, bare=False):
-        if not os.path.isdir(self.repo_path):
-            os.makedirs(self.repo_path, exist_ok=True)
+        if not os.path.isdir(self.path):
```
danixeee: `self._path.is_dir()`

taf/git.py
```diff
@@ -475,8 +465,8 @@ def get_tracking_branch(self, branch=""):
...
+            os.makedirs(self.path, exist_ok=True)
```
danixeee: `self._path.mkdir(exists=True, parents=True)`

taf/repository_tool.py
```diff
@@ -135,30 +143,36 @@ def _check_key_and_get_pin(expected_key_id):
 
 
 class Repository:
-    def __init__(self, repository_path, repository_name="default"):
-        self.repository_path = repository_path
+    def __init__(self, repo_path, repo_name="default"):
+        self.path = repo_path
```
danixeee: `self._path = Path(repo_path` + property `path` maybe makes sense here as well.

taf/git.py
renatav: I think that that makes sense. I think that one the reason why we couldn't use `pathlib`'s path was because of `Python 3.5` issues. Since we decided not to support `Python 3.5`, we can transition to using `pathlib`'s path where that is possible. We could keep that in mind while working on other issues. I add this now though

PR #114, opened 2020-03-23
Create repository improvements and some fixes

### LINKED ISSUE #95: Create an empty targets directory automatically if it does not already exist
When executing certain commands, we need to instantiate `tuf` repository. However, that is not possible if there is no `targets` directory

### LINKED ISSUE #108: Create repository further improvements
- When creating a new repository, if configuration json is not used and the user has to answer all of the setup questions, output the user's answers as a json
- When creating a new repository, do not raise an error if the directory exists - just in case when a true authentication repository is already there
- When a key cannot be loaded 
  - output the location that is trying to be loaded
  - make it clear if private or public key is missig/invalid
  - if key is corrupt, print that as the error message
- Add keystore to keys_description json as an optional parameter
- Handle existing `repositories.json`

renatav:

`keys_description.json` is now expected to contain a "roles" dictionary and a "keystore" location (though both are optional), like this:
```
{
"roles": {
  "root": {
    "number": 2,
    "length": 2048,
	"threshold": 2
  },
  "targets": {
    "length": 2048
  },
  "snapshot": {},
  "timestamp": {}
  },
  "keystore": "keystore_path"
  
}
```

Update snapshot and timestamp when adding a new signing key
Create repository improvements

Closes #108 
Closes #95 

### DIFF
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -443,13 +443,13 @@ def _initialize_roles_and_keystore(roles_key_infos, keystore, enter_info=True):
     Read or enter roles information and try to extract keystore path from
     that json
     """
-    roles_key_infos = read_input_dict(roles_key_infos)
-    if enter_info and not len(roles_key_infos):
+    roles_key_infos_dict = read_input_dict(roles_key_infos)
+    if enter_info and not len(roles_key_infos_dict):
         # ask the user to enter roles, number of keys etc.
-        roles_key_infos = _enter_roles_infos(keystore)
+        roles_key_infos_dict = _enter_roles_infos(keystore, roles_key_infos)
     if keystore is None:
-        keystore = roles_key_infos.get("keystore")
-    return roles_key_infos, keystore
+        keystore = roles_key_infos_dict.get("keystore")
+    return roles_key_infos_dict, keystore
 
 
 def _setup_roles_keys(
@@ -574,7 +574,7 @@ def _setup_yubikey(yubikeys, role_name, key_name, scheme, certs_dir):
         return key
 
 
-def _enter_roles_infos(keystore):
+def _enter_roles_infos(keystore, roles_key_infos):
     """
     Ask the user to enter information taf roles and keys, including the location
     of keystore direcotry if not entered through an input parameter
@@ -584,7 +584,7 @@ def _enter_roles_infos(keystore):
     infos_json = {}
 
     for role in mandatory_roles:
-        role_key_infos[role] = _enter_role_info(role, role == "targets")
+        role_key_infos[role] = _enter_role_info(role, role == "targets", keystore)
     infos_json["roles"] = role_key_infos
 
     while keystore is None:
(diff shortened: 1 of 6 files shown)


taf/developer_tool.py
```diff
@@ -268,16 +298,20 @@ def create_repository(
         Indicates if the created repository is a test authentication repository
     """
     yubikeys = defaultdict(dict)
+    auth_repo = AuthenticationRepo(repo_path)
+    repo_path = Path(repo_path)
+
+    if not _check_if_can_create_repository(auth_repo):
```
danixeee: Could you just call `auth_repo._tuf_repository` and if it raises an error, you will know there there is no (or invalid) tuf repository at the path.

taf/developer_tool.py
```diff
@@ -44,7 +45,12 @@
 
 
 def add_signing_key(
-    repo_path, role, pub_key_path=None, scheme=DEFAULT_RSA_SIGNATURE_SCHEME
+    repo_path,
+    role,
+    pub_key_path=None,
+    keystore=None,
+    roles_key_infos=None,
+    scheme=DEFAULT_RSA_SIGNATURE_SCHEME,
```
danixeee: docstring should be updated

taf/developer_tool.py
```diff
@@ -95,6 +107,7 @@ def add_signing_key(
             num_of_signatures += 1
         if num_of_signatures == keys_num:
             all_loaded = True
+    update_roles(["snapshot", "timestamp"], taf_repo, keystore, role_infos)
```
danixeee: `keystore` could still be `None` here and I guess it will fail? This is fine for now, but when it comes to key rotation, I am not sure who will poses _snapshot_ and _timestamp_ keys. Probably we would use CI to sign those roles. I suggest  checking if `keystore is None` and skipping `update_roles` in that case, with a proper warning message.

taf/developer_tool.py
```diff
@@ -429,26 +474,42 @@ def _setup_keystore_key(
     # if keystore exists, load the keys
     generate_new_keys = keystore is None
     public_key = private_key = None
+
+    def _invalid_key_message(key_name, keystore, is_public):
+        extension = ".pub" if is_public else ""
+        key_path = Path(keystore, f"{key_name}{extension}")
+        if key_path.is_file():
+            print(f"{key_path} is not valid!")
```
danixeee: f"{key_path} is not a file!" sound better to me

taf/developer_tool.py
```diff
@@ -969,6 +1063,11 @@ def _write_targets_metadata(taf_repo, targets_roles, keystore, roles_key_infos,
             update_method = taf_repo.roles_keystore_update_method(role_name)
             update_method(keystore_keys, write=False)
 
+
+def _write_targets_metadata(taf_repo, targets_roles, keystore, roles_key_infos, scheme):
```
danixeee: Name of the function is `_write_targets_metadata`, but it updates "snapshot", "timestamp" as well. As in previous comment, I am not sure if those should always be signed. Does this affect a build process where just target metadata should be generated?

taf/developer_tool.py
renatav: This is meant to check if repository already exists. If the directory contains a git repository with metadata files, reinitializing metadata files could result in an invalid repo (e.g. versions of metadata being reverted to 1)

taf/developer_tool.py
renatav: If `keystore is None` and these keys are not stored on Yubikeys, the users will be asked to directly enter these keys. This is handled by _load_signing_keys called from update_roles. Not sure if we'll need to modify this once we need to handle key rotation, but there won't be an error if `keystore is None` for now

taf/developer_tool.py
renatav: I think that it's fine because we always want to sign snapshot and timestamp in this case. `developer_tool` is not used by our build process

renatav:
@danixeee 
1. I don't think that we should automatically save the file to authentication repository as that might not be something we want to commit. Instead, if `--keys-description` is a string and the configuration file does not already exist at that location,we will attempt to write this json there. In other cases, we will print it
2. That makes sense, but after looking into it for a bit, I am not sure that there is an easy way to get key length given a keystore file. The dictionary containing information about the keys upon reading them does not contain this information and I couldn't find anything useful in `securesystems` lib. Not sure that it's worth adding another dependency just to do that. If you've seen a way of getting that length for depending on the scheme and knowing private and/or public key, I can implement that now. But I vote for not spending to much time on that as the end users won't use this I believe. We could create an issue for this.
3. Implemented

danixeee:
1. Sound good to me
2. It's easy with `cryptography` library which is already required for signing:

```python
def get_key_size(key_pem_path, decrypt_pwd=None):
    from cryptography.hazmat.backends import default_backend
    from cryptography.hazmat.primitives.serialization import load_pem_public_key, load_pem_private_key

    try:
        key = load_pem_public_key(Path(key_pem_path).read_bytes(), default_backend())
    except ValueError:
        key = load_pem_private_key(Path(key_pem_path).read_bytes(), decrypt_pwd, default_backend())
    except Exception:
        return 0

    return key.key_size
```

PR #120, opened 2020-04-06
Various bug fixes

### LINKED ISSUE #109: ykpers error message should not appear
Do not show that error message unless the user attempts to use a yubikey

### LINKED ISSUE #119: Keystore signing fixes
- [x] If a user wants to reuse an exiting key pair loaded from keystore when creating a new repository, automatically create public and private keys whose content is identical to the loaded ones, with name which is derived from the current role's name
- [x] When loading signing keys, if there are additional keys which should not be loaded in the keystore, do not attempt to load them. For example, if number of keys is of the targets role is 1, and there are `targets` and `targets.pub`, but also `targets1` and `targets1.pub`, do not try to load the second pair.

danixeee:

Closes #119.
Closes #109.

### DIFF
--- a/taf/developer_tool.py
+++ b/taf/developer_tool.py
@@ -17,7 +17,7 @@ from tuf.repository_tool import (
     generate_rsa_key,
 )
 
-from taf import YUBIKEY_MANAGER_ERR_MSG
+from taf import YubikeyMissingLibrary
 from taf.auth_repo import AuthenticationRepo
 from taf.constants import DEFAULT_RSA_SIGNATURE_SCHEME
 from taf.exceptions import KeystoreError, TargetsMetadataUpdateError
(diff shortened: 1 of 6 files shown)


taf/developer_tool.py
```diff
@@ -466,11 +469,13 @@ def _initialize_roles_and_keystore(roles_key_infos, keystore, enter_info=True):
     that json
     """
     roles_key_infos_dict = read_input_dict(roles_key_infos)
+    if keystore is None:
+        keystore = roles_key_infos_dict.get("keystore", _default_keystore_path())
```
renatav, line 473: When `keystore` is specified in `roles_key_infos` (in configuration json file), that path will be used, but the line saying that the default location will be used is still printed

PR #121, opened 2020-04-10
Git updates

renatav:

- Made clone more flexible - added an option to pass in additional arguments. Useful if e.g. we need a shallow clone
- Fixed `synced_with_remote`. We used to only check if local and remote tracking branch have the same top commit. This does not work if the remote tracking branched is not up to date.

### DIFF
--- a/taf/git.py
+++ b/taf/git.py
@@ -509,7 +509,7 @@ class GitRepository:
 
     def get_last_remote_commit(self, url, branch="master"):
         """
-        get the last remote commit without cloning the repo
+        Fet the last remote commit of the specified branch
         """
         if url is not None:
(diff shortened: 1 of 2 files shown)


taf/git.py
```diff
@@ -290,18 +290,24 @@ def checkout_paths(self, commit, *args):
...
+        params = []
         if bare:
-            params = "--bare"
+            params.append("--bare")
         elif no_checkout:
-            params = "--no-checkout"
+            params.append("--no-checkout")
+
+        for name, value in kwargs.items():
+            params.append(f"--{name} {value}")
```
danixeee, line 307: If we have `option=True/False` in `**kwargs`, then I think you should only append a `name` and not a `value`.

PR #131, opened 2020-05-04
Error handling and logging improvements

renatav:

- Standardize log and error message
- Added a `GitError` exception which is raised instead of reraising `CalledProcessError` error
- Went through git methods and updated logging
- Raise a custom error if a repository cannot be instantiated
- Small updater improvements - call cleanup even if inistialization fails (and not just if update fails)

### DIFF
--- a/taf/repositoriesdb.py
+++ b/taf/repositoriesdb.py
@@ -98,11 +98,11 @@ def load_repositories(
             repositories = _get_json_file(auth_repo, repositories_path, commit)
             targets = _get_json_file(auth_repo, targets_path, commit)
         except InvalidOrMissingMetadataError as e:
-            log_func = taf_logger.warning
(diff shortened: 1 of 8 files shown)


taf/repositoriesdb.py
```diff
@@ -94,7 +98,10 @@ def load_repositories(
             repositories = _get_json_file(auth_repo, repositories_path, commit)
             targets = _get_json_file(auth_repo, targets_path, commit)
         except InvalidOrMissingMetadataError as e:
-            taf_logger.warning("Skipping commit {} due to error {}", commit, e)
+            log_func = taf_logger.warning
```
danixeee, line 101: Could we just log with _debug_ level? Not sure what benefits we have with separating these logs, but I would rather raise some `RepositoriesJsonMissingError` exception and catch it here if needed.

PR #161, opened 2020-12-12
Various updater fixes and enhancements

renatav:

- Raise an error if there are additional commits newer than the last authenticated commit if the updater is called with the `check-authenticated` flag 
- A minor validation command fix
- Some improvements made to the updater's error messages
- Initial worktrees support . When merging commits, we need to check out branches which need to be updated. That cannot be done if they are checked out in a different worktree. So try to switch to that worktree and update the branch. Currently, there is no validation if that is possible or permissible. 
- Added support for specifying location of the conf directory
- Replaced authenticate-test-repo flag with an enum
- Added a function for disabling fie logging

### DIFF
--- a/taf/exceptions.py
+++ b/taf/exceptions.py
@@ -131,15 +131,16 @@ class UpdateFailedError(TAFError):
 
 class UpdaterAdditionalCommits(TAFError):
     def __init__(self, additional_commits_per_repo, message=None):
-        self.repositories = list(additional_commits_per_repo.keys())
+        self.additional_commits_per_repo = additional_commits_per_repo
         if message is None:
             message = ''
-            for repo, commits in additional_commits_per_repo.items():
-                message += f"Repository {repo} contains {len(commits)} additional"
-                if len(commits) > 1:
-                    message += "commits\n"
-                else:
-                    message += "commit\n"
+            for repo, branch_commits in additional_commits_per_repo.items():
+                for branch, commits in branch_commits.items():
+                    message += f"Repository {repo}: branch {branch} contains {len(commits)} additional"
+                    if len(commits) > 1:
+                        message += "commits\n"
+                    else:
+                        message += "commit\n"
         self.message = message
 
 
--- a/taf/exceptions.py
+++ b/taf/exceptions.py
@@ -129,5 +129,19 @@ class UpdateFailedError(TAFError):
     pass
 
 
+class UpdaterAdditionalCommits(TAFError):
+    def __init__(self, additional_commits_per_repo, message=None):
+        self.repositories = list(additional_commits_per_repo.keys())
+        if message is None:
+            message = ''
+            for repo, commits in additional_commits_per_repo.items():
+                message += f"Repository {repo} contains {len(commits)} additional"
+                if len(commits) > 1:
+                    message += "commits\n"
+                else:
+                    message += "commit\n"
+        self.message = message
+
+
 class YubikeyError(Exception):
     pass
--- a/taf/exceptions.py
+++ b/taf/exceptions.py
@@ -133,7 +133,7 @@ class UpdaterAdditionalCommits(TAFError):
     def __init__(self, additional_commits_per_repo, message=None):
         self.additional_commits_per_repo = additional_commits_per_repo
         if message is None:
-            message = ''
+            message = ""
             for repo, branch_commits in additional_commits_per_repo.items():
                 for branch, commits in branch_commits.items():
                     message += f"Repository {repo}: branch {branch} contains {len(commits)} additional"
--- a/taf/exceptions.py
+++ b/taf/exceptions.py
@@ -129,7 +129,7 @@ class UpdateFailedError(TAFError):
     pass
 
 
-class UpdaterAdditionalCommits(TAFError):
+class UpdaterAdditionalCommitsError(TAFError):
     def __init__(self, additional_commits_per_repo, message=None):
         self.additional_commits_per_repo = additional_commits_per_repo
         if message is None:
--- a/taf/git.py
+++ b/taf/git.py
@@ -609,6 +609,14 @@ class GitRepository:
                 branch = ""
             self._git("fetch {} {}", remote, branch, log_error=True)
 
+    def find_worktree_path_by_branch(self, branch_name):
+        """Returns worktree path where branch is checked out."""
+        worktrees = self.list_worktrees()
+        for path, _, _branch_name in worktrees.values():
+            if _branch_name == branch_name:
+                return path
+        return None
+
     def get_current_branch(self):
         """Return current branch."""
         return self._git("rev-parse --abbrev-ref HEAD").strip()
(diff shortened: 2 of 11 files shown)


taf/updater/updater.py
```diff
@@ -345,9 +353,14 @@ def _update_target_repositories(
             if is_git_repository:
                 repository.fetch(branch=branch)
             if old_head is not None:
-                new_commits_on_repo_branch = repository.all_fetched_commits(
-                    branch=branch
-                )
+                if not only_validate:
```
dgreisen: could we reverse and just say `if only_validate`? using not and else makes my head hurt

taf/updater/updater.py
```diff
@@ -432,27 +432,38 @@ def _update_target_repositories(
...
+                    repository,
+                    branch,
+                    branch_commits,
+                    allow_unauthenticated[path],
+                    new_commits[path][branch],
+                )
     return additional_commits_per_repo
 
 
+def _merge_branch_commits(
```
dgreisen, line 504: docstring

taf/updater/updater.py
```diff
@@ -445,6 +424,38 @@ def _update_target_repositories(
     return additional_commits_per_repo
 
 
+def _get_commits(
```
dgreisen, line 458: docstring

taf/git.py
```diff
@@ -609,6 +609,14 @@ def fetch(self, fetch_all=False, branch=None, remote="origin"):
                 branch = ""
             self._git("fetch {} {}", remote, branch, log_error=True)
 
+    def find_worktree_path_by_branch(self, branch_name):
+        """Returns worktree path where branch is checked out."""
```
dgreisen, line 613: suggestion: `, or None if not checked out in any worktree`

taf/updater/updater.py
```diff
@@ -431,7 +436,14 @@ def _get_commits(
         repository.fetch(branch=branch)
```
dgreisen, line 462: docstring for _get_commits

taf/updater/updater.py
```diff
@@ -459,18 +476,36 @@ def _get_commits(
...
-    taf_logger.info("Merging {} into {}", last_commit, repository.name)
     last_validated_commit = last_commit
     commit_to_merge = (
         last_validated_commit if not allow_unauthenticated else new_branch_commits[-1]
     )
+    taf_logger.info("Merging {} into {}", commit_to_merge, repository.name)
+    _merge_commit(repository, branch, commit_to_merge, allow_unauthenticated)
+
+
+def _merge_commit(repository, branch, commit_to_merge, allow_unauthenticated=False):
```
dgreisen, line 516: docstring

taf/tools/repo/__init__.py
```diff
@@ -189,7 +189,9 @@ def initialize(path, root_dir, namespace, targets_rel_dir, custom, use_mirrors,
                   "repository from the filesystem")
     @click.option("--authenticate-test-repo", is_flag=True, help="Indicates that the authentication "
                   "repository is a test repository")
-    def update(url, clients_auth_path, clients_root_dir, from_fs, authenticate_test_repo):
+    @click.option("--check-for-unauthenticated", is_flag=True, help="Check if there are additional new commits "
```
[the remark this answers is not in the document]
dgreisen: do we really want to set default to either? Wouldn't it be safer to set default to official?

dgreisen (changes requested):

taf/exceptions.py
```diff
@@ -129,5 +129,19 @@ class UpdateFailedError(TAFError):
     pass
 
 
+class UpdaterAdditionalCommits(TAFError):
```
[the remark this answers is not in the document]
renatav: By conventions, all TAF exceptions should end with "Error", so totally agree, will rename

taf/repository_tool.py
```diff
@@ -65,6 +65,18 @@ def is_delegated_role(role):
     return role not in ("root", "targets", "snapshot", "timestamp")
 
 
+def is_auth_repo(repo_path):
```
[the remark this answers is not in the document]
renatav: We rely on TUF to check if the folder contains a valid repository. TUF raises an error if the metadata directory exists, but targets does not. I am not sure when exactly we have the case of metadata directory existing and targets not, that instantiation of TUF's repository class is among oldest parts of TAF code. It would make sense to make sure that we are generating both metadata and targets directories when, say, creating a new authentication repository using TAF's cli. Since we do not have developer tool tests yet, I would prefer not to touch that code for now. I think that it makes sense to move removal of targets directory to the `except` block of the method which creates it if it does not exist.

taf/tools/repo/__init__.py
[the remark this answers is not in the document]
renatav: `error-if-unauthenticated` is probably closer to what we are doing here. I don't have a strong opinion regarding the default value. Setting the default value to `official` would basically be the same thing as having the `authenticate-test-repo` flag and omitting it when calling `update` (which would result in successful update of official repositories seeing that they do not contain any flags)

taf/git.py
```diff
@@ -722,6 +730,14 @@ def list_untracked_files(self, path=None):
             untracked_file for untracked_file in untracked_files if len(untracked_file)
         ]
 
+    def list_worktrees(self):
+        worktrees_list = self._git("worktree list")
```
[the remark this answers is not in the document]
dgreisen: docstring should include what is being returned. It looks like it's actually a dict where the key is a tuple with data unknown unless I run `worktree list` myself and figure out the output

taf/tools/repo/__init__.py
```diff
@@ -212,10 +213,15 @@ def update(url, clients_auth_path, clients_root_dir, from_fs, expected_repo_type
         the "test" target file), use --authenticate-test-repo flag. An error will be raised
         if this flag is omitted in the mentioned case. Do not use this flag when validating a non-test
         repository as that will also result in an error.
+
+        Some repositories can contain unauthenticated commits in-between two authenticated. The updater
+        will check if existance and order of all authenticated ones and ignore unauthenticated. To raise
+        an error if there new authenticated commits (newer than the last local commit), error-if-unauthenticated
```
dgreisen, line 219: question: should this be "error if there **are** new **un**authenticated commits"?

taf/updater/updater.py
```diff
@@ -514,6 +519,8 @@ def _merge_branch_commits(
 
 
 def _merge_commit(repository, branch, commit_to_merge, allow_unauthenticated=False):
+    """ Merged the specified branch into the given branch
```
dgreisen, line 522: Merge~d~ the specified **commit** into the given branch? what does allow_unauthenticated mean? is that the same as error-if-unauthenticated?

dgreisen (changes requested):

taf/tools/repo/__init__.py
renatav: Yes

taf/updater/updater.py
renatav: `allow_unatuhenticated` indicates if a repository can contain authenticated commits and is specified in the authenticated repository. I think that it will be necessary to put more effort into figuring out what exactly to check out at the end of update process. At the moment, repositories where unauthenticated commits are allowed and where they are not are handled differently - but this decision was based on my analysis of our needs more so than anything else.

taf/git.py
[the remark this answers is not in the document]
dgreisen: excellent!

dgreisen (approved):

PR #213, opened 2022-03-02
Optional clients auth path, validation rework

### LINKED ISSUE #214: Optional clients-auth-path
We need to determine `auth_repo_name` if clients-auth-path isn't provided. Make clients-auth-path an optional argument.

n-dusan:

Closes #214 

New use case: we should be able to use updater if only `clients-library-dir` and `urls` (without `clients-auth-path`) are provided. Approach:
*  Initially, auth_repo_name is None. 
* Determine auth_repo_name during validation phase by checking validation temp dir and
  extracting {namespace}/{name}. We expect a signed `targets/protected/info.json` which provides repository metadata.

Temp dir validation is now moved to `updater.py`.
Add a new settings field 


taf/updater/updater.py
```diff
@@ -68,6 +68,39 @@ def _check_update_status(repos_update_data, auth_repo_name, host_update_status,
...
+    validation_auth_repo.fetch(fetch_all=True)
+
+    settings.validation_repo_path = validation_auth_repo.path
+
+    validation_head_sha = validation_auth_repo.top_commit_of_branch(default_branch)
+
+    if repository_name is None:
+        try:
+            info = validation_auth_repo.get_json(
+                validation_head_sha, "targets/protected/info.json"
```
renatav, line 92: info.json path should be a constant. At some point, we should move all of the paths that the framework needs to the same place

taf/updater/handlers.py
```diff
@@ -237,25 +236,23 @@ def _init_metadata(self):
             current_filename.write_text(metadata)
             shutil.copyfile(str(current_filename), str(previous_filename))
 
-    def _clone_validation_repo(self, url):
+    def _set_validation_repo(self, path, url):
```
renatav, line 239: Should not start with _ if accessed from outside of the module

taf/updater/updater.py
```diff
@@ -564,8 +614,13 @@ def _update_current_repository(
     tuf.settings.repositories_directory = clients_auth_library_dir
 
     # check whether the directory that runs the cloning exists or contains additional files.
-    # we need to check the state of folder before running tuf.
-    users_repo_existed = Path(clients_auth_library_dir, auth_repo_name).exists()
+    # we need to check the state of folder before running tuf. Resolves issue #22
+    # if auth_repo_name isn't specified then the current directory doesn't contain additional files.
+    users_repo_existed = (
+        Path(clients_auth_library_dir, auth_repo_name).exists()
+        if auth_repo_name is not None
```
renatav, line 621: Don't thnk that we actually can know if the clients folder exists if we don't know the path

renatav (commented):
Looks good, left minor comments. Please only run taf repo validate to make sure it is still working

renatav (approved):

PR #245, opened 2022-05-05
Check for YubiKey roles in `_unchanged_targets metadata`

n-dusan:
* If inserted yubikey roles exist, suppose they are changed and remove
  them from the list of unchanged roles.

If a YubiKey is inserted, extract the roles from it and consider those roles changed. They won't be included in `all_roles`.

### DIFF
--- a/taf/validation.py
+++ b/taf/validation.py
@@ -5,6 +5,7 @@ from tuf.repository_tool import TARGETS_DIRECTORY_NAME, METADATA_DIRECTORY_NAME
 from taf.repository_tool import Repository
 from taf.constants import CAPSTONE
 from taf.exceptions import InvalidBranchError
+from taf.yubikey import is_inserted, get_piv_public_key_tuf
 
 
 def validate_branch(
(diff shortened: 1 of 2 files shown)


taf/validation.py
```diff
@@ -197,5 +198,12 @@ def _compare_commit_with_targets_metadata(
 def _get_unchanged_targets_metadata(auth_repo, updated_role):
     taf_repo = Repository(auth_repo.path)
     all_roles = taf_repo.get_all_targets_roles()
-    all_roles.remove(updated_role)
+    yk_inserted = is_inserted()
+    updated_roles = [updated_role]
+    if yk_inserted:
```
renatav, line 203: I think that this might be coupling our specific needs to closely with this validation (which is supposed to be as generic as possible while still covering out requirements). Can this function receive a list of updated roles instead, so that the YubiKey logic is fully in our closed source code?

n-dusan:
Closed in favor #246

taf/validation.py
n-dusan: Thanks, list of updated roles are now in #246

PR #259, opened 2022-07-18
Fix cleanup error messages after failed update

### LINKED ISSUE #243: Cleanup error/error message after failed update
When running the updater for the first time (cloning repositories) and something goes wrong, a cleanup will be attempted. This cleanup often raises an error e.g `"[WinError 5] Access is denied: 'D:\\\\webnode\\\\open-law-library\\\\library-dev\\\\mohicanlaw\\\\law\\\\.git\\\\objects\\\\pack\\\\pack-bf154c865f4fce50647863222a457968f8f6d506.idx'"}`

This error is then raised by the update process instead of the original error - why the validation failed in the first place

n-dusan:
* Basing on #243, we are having some difficulty with pygit2 Repository file
  handlers being open. in `pygit.py` we instantiate the Repository
  class, which holds references to `pack` and `.idx` git objects and
  keeps files open. If taf and cleanup is attempted from same subprocess,
  Repository handler won't be deleted, and cleanup error occurs. Calling
  manual garbage collection did not help. This
  seems to be a known issue throughout the git community, see [1] and
  [2].

  [1] - https://github.com/libgit2/pygit2/issues/596
  [2] - https://github.com/gitpython-developers/GitPython/issues/546

  Several workarounds were proposed by the community, but it seems that
  we can't fully replicate those workarounds. In [2] they attempt to
  workaround the issue by deleting git cache. That works for them
  because their git implementation has file handlers open in pure
  python, while we only do subprocess git calls.

  Our TODO: if these  windows specific error issues keep persisting, figure out how to
  garbage collect pygit2 Repository handler correctly.

  For now settle on error handling, as this issue is
  specific to a subprocess call that calls taf/updater. Though we should
  keep track of progress of these known bugs, so we can fix them if they
  get addressed by pygit2 or gitpython

Closes #243 

Before cleanup/error-message with `--format-output` on an expired partner::  
```
{"updateSuccessful": false, "error": "[WinError 5] Access is denied: 'C:\\\\OLL\\\\platform\\\\integration-tests\\\\oll\\\\integration_tests\\\\build_and_merge\\\\repositories\\\\test_build_speculative_with_updater\\\\cityofupdatertests\\\\cityoftests\\\\law-html\\\\.git\\\\objects\\\\pack\\\\pack-1da7256d4aebdc6770807a79948210beb36ac5a0.idx"}
```
After:
```
{"updateSuccessful": false, "error": "Validation of authentication repository XX/law failed at revision 9d571c1c5753009b82cafdceb63dc7a6d1699da0 due to error: No working mirror was found:\n  '': ExpiredMetadataError(\"Metadata 'timestamp' expired on Sat Feb 12 16:29:01 2022 (UTC).\")Failed to cleanup repository updater: [WinError 5] Access is denied: 'C:\\\\OLL\\\\platform\\\\integration-tests\\\\oll\\\\integration_tests\\\\build_and_merge\\\\repositories\\\\test_build_speculative_with_updater\\\\cityofupdatertests\\\\cityoftests\\\\law-html\\\\.git\\\\objects\\\\pack\\\\pack-1da7256d4aebdc6770807a79948210beb36ac5a0.idx"}
```


taf/updater/handlers.py
```diff
@@ -257,7 +257,15 @@ def cleanup(self):
         self.validation_auth_repo.cleanup()
         temp_dir = Path(self.validation_auth_repo.path, os.pardir).parent
         if temp_dir.is_dir():
-            shutil.rmtree(str(temp_dir), onerror=on_rm_error)
+            try:
+                shutil.rmtree(str(temp_dir), onerror=on_rm_error)
+            except (OSError, PermissionError) as e:
+                taf_logger.error(
```
renatav, line 263: Thinking that this is more of a warning than an error. If there were no other issues, but cleanup failed, that should not result in an error, right? (I know that the broader context is that we have issues when an error occurs)

taf/updater/updater.py
```diff
@@ -1305,22 +1305,25 @@ def _validate_authentication_repository(
...
-            error_msg = UpdateFailedError(
+            error_msg = (
                 f"Repository {users_auth_repo.name} is not a test repository,"
                 ' but update was called with the "--expected-repo-type" test'
             )
 
     # always cleanup repository updater
-    repository_updater.update_handler.cleanup()
+    try:
+        repository_updater.update_handler.cleanup()
```
renatav, line 1321: This is called even when no errors happen? So if cleanup fails, should update as a whole raise an error? Without this fix, update just fails, so this certainly is not a regression. Should this be a warning like when pip prints that we should upgrade?

renatav (commented):
Some thoughts, but looks good. We can discuss these points, but will approve

renatav (approved):

taf/updater/handlers.py
n-dusan: Makes sense, moved message to `on_rm_error` handler as a warning

PR #260, opened 2022-07-19
`--clients-auth-path` repo command improvements

### LINKED ISSUE #217: Improve --clients-auth-path error message
When running the updater without `--clients-auth-path` flag, the output is:
`*** taf.exceptions.GitError: Repo tmpoe1dtbhe/auth_repo: fatal: Path 'targets/protected/info.json' does not exist in 'f62502e78a14995c3dd8dbc528321c5551a78f10'`

n-dusan:

Closes #217 

- Revert `taf repo update` click command back to old API by having `--clients-auth-path` as an argument. Now that argument is optional. `taf repo update` can be run by either specifying `<url>` and `<clients-auth-path>`, or with `<url>` and `--clients-library-dir`
- Do shutil.rmtree after unsuccessful `taf repo update` but not during validate. We don't want to clean up a law repository if validate fails.
- Clarify `info.json` error message

### DIFF
--- a/taf/tools/repo/__init__.py
+++ b/taf/tools/repo/__init__.py
@@ -181,7 +181,7 @@ def attach_to_group(group):
 
     @repo.command()
     @click.argument("url")
-    @click.option("--clients-auth-path", default=None, help="Directory where authentication repository is located.")
+    @click.argument("clients-auth-path", default=None, required=False)
     @click.option("--clients-library-dir", default=None, help="Directory where target repositories and, "
                   "optionally, authentication repository are located. If omitted it is "
                   "calculated based on authentication repository's path. "
--- a/taf/tools/repo/__init__.py
+++ b/taf/tools/repo/__init__.py
@@ -209,12 +209,16 @@ def attach_to_group(group):
         locations of the target repositories are calculated based on the authentication repository's
         path. If that is not the case, it is necessary to redefine this default value using the
         --clients-library-dir option. This means that if authentication repository's path is
-        E:\\root\\namespace\\auth-repo, it will be assumed that E:\\root is the root direcotry
+        E:\\root\\namespace\\auth-repo, it will be assumed that E:\\root is the root directory
         if clients-library-dir is not specified.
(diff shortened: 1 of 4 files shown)


taf/tools/repo/__init__.py
```diff
@@ -181,7 +181,7 @@ def initialize(path, library_dir, namespace, targets_rel_dir, custom, use_mirror
 
     @repo.command()
     @click.argument("url")
-    @click.option("--clients-auth-path", default=None, help="Directory where authentication repository is located.")
+    @click.argument("clients-auth-path", default=None, required=False)
```
renatav, line 184: The reason why we tend to use options is because it allows us to use the help parameter, which results in nice `--help` output. I can see why it would be nice to lose `--`, but then we need to update the docstring and explain clients-auth-path's purpose

taf/updater/updater.py
```diff
@@ -739,7 +739,7 @@ def _commits_ret(commits, existing_repo, update_successful):
         )
     except Exception as e:
 
-        if not existing_repo:
+        if not existing_repo and not only_validate:
```
renatav, line 742: Feels like something's wrong here if it's possible for `existing_repo` can be `None` and `only_validate` `True`. We only validate existing repos. What is deleting existing repos?

n-dusan: Correct, removing that

taf/tools/repo/__init__.py
n-dusan: Thought it makes sense to transition taf back to old implementation that is similar to how `git clone` works. `git clone` has a URL and a second (optional) argument. The second argument made it possible to clone to current directory. So for example `git clone https://github.com/openlawlibrary/taf .` clones to current directory without specifying any flags. What do you think?

renatav: Yeah, that's easier to use, but we should make sure that docstring is updated following removal of help.

renatav (approved):
Looks good

PR #276, opened 2022-10-26
Downgrade expired metadata to a warning

### LINKED ISSUE #272: Make Updater not fail if metadata expires
Show warning if targets have expired.
- [x] Add --error-if-warning flag to revert back to current TAF state.

n-dusan:
Closes #272 

Added a `--error-if-warning` flag to indicate if TAF should fail on warnings or not. I've seen that other frameworks also call that a "strict" mode. By default, I think we want to turn strict mode off, and opt into strict mode by adding `--error-if-warning`. 
For now, only point of failure is the updater expired metadata, which we now turn into a warning.

Seems much simpler to just set a `settings` configuration for strict mode, and turn it on/off when `--error-if-warnings` is set.


taf/settings.py
```diff
@@ -16,6 +16,10 @@
 
 default_branch = "main"
 
+# Strict mode enabled/disabled. If error_if_warning is enabled, any warnings
+# should raise TAF errors
+error_if_warning = False
```
renatav, line 21: I actually like the name `strict` a lot more. `error_if_warning` sounds contradictory - if it's a warning, it's a warning, if it's an error, it's an error.

taf/updater/updater.py
```diff
@@ -797,20 +802,23 @@ def _update_authentication_repository(repository_updater):
...
-            current_commit,
-            e,
-        )
-
-        raise UpdateFailedError(
-            f"Validation of authentication repository {users_auth_repo.name}"
-            f" failed at revision {current_commit} due to error: {e}"
-        )
+        except Exception as e:
+            metadata_expired = EXPIRED_METADATA_ERROR in type(
```
renatav, line 806: `ExpiredMetadataError` is not raised in the newest version of the updater that we just merged in, so I don't think that this will work after merging `master`. Would it be easier to handle everything on the handler's side?

renatav (changes requested):
I think that this will have to be updated now that we updated our TUF fork. Changes made to tests and everything else look good to me

taf/settings.py
n-dusan: That makes sense, renamed `error_if_warning` to `strict`

taf/updater/updater.py
n-dusan: Actually, it does raise! Not very different error handling than the old TUF.

renatav (approved):

PR #292, opened 2022-11-25
Branch check update - support first commits on branches with a missing branch file

renatav:

In some cases, branch file might not be present in the first new commit
of the authentication repository. check_branch_roles turned into a dictionary
where repo name is the key and a flag indicating if a missing branch
is a valid state or not in case of that repo. This should be made
more generic in the future

A branch file can be inside a folder named after the role inside the
targets folder, but also directly inside targets, depending on
how many roles need these files. Only try to load a branch file
from a certain location if that location was delegated to the current role

### DIFF
--- a/taf/validation.py
+++ b/taf/validation.py
@@ -125,20 +125,17 @@ def _check_lengths_of_branches(targets_and_commits, branch_name):
 
 def _check_branch_id(auth_repo, auth_commit, branch_id, role, is_first_commit):
 
-    try:
-        new_branch_id = auth_repo.get_file(
-            auth_commit, Path(TARGETS_DIRECTORY_NAME, role, "branch")
-        )
-    except GitError:
-        try:
-            new_branch_id = auth_repo.get_file(
-                auth_commit, Path(TARGETS_DIRECTORY_NAME, "branch")
-            )
-        except GitError:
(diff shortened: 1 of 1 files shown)


taf/validation.py
```diff
@@ -90,8 +91,14 @@ def validate_branch(
...
             if updated_role in check_branch_roles:
+                no_initial_branch_id = check_branch_roles[updated_role]
                 branch_id = _check_branch_id(
-                    auth_repo, auth_commit, branch_id, updated_role
+                    auth_repo,
+                    auth_commit,
+                    branch_id,
+                    updated_role,
+                    is_first_commit=no_initial_branch_id
+                    and commit_index == len(auth_commits) - 1,
```
n-dusan, line 101: Approved. Minor comment, non blocking: would it maybe be simpler if we skip calling `_check_branch_id`  in case of `is_first_commit` being True?

n-dusan (approved):

taf/validation.py
renatav: That makes sense, but what we really want to do is implement this more generically. I dislike this solution, this is the part of the code that I kept implementing and losing due to hard resets and I would like to come up with a better way of handling this

Issue #317, opened 2023-04-10
Remove remaining functionalities from `developer_tool`


renatav:
Functions which are still in `developer_tool` all use `roles_key_infos`, expecting a `json` file containing information about the repository's roles. The question is if we want to keep using this `json` file. If this file is not provided the commands still work, but it's necessary to enter additional info. It makes sense to check if which data can we extracted from authentication repository's metadata files instead - unless this file is used to simplify commands like creation of a new repository and generation of new keystore file. When adding new roles, it should not be necessary to provide a file which contains information about all existing roles that should not be updated.
Once other functions using `roles_key_infos` are reworked, they should be moved to appropriate api modules. `developer_tool` shoud be removed.

- #159  can also be resolved as a part of cli refactroring and removal of `developer_tool`

PR #320, opened 2023-05-06
Try to sign using a yubikey before asking the user if they want to use a yubikey

### LINKED ISSUE #315: Try to sign using a yubikey before asking the user if they want to use a yubikey
Try signing with a yubikey first, then ask if they want to use one or a keystore file if that is not successful

renatav:

Rework `load_signing_keys`. Try to sign using a yubikey before asking the user if they want to use a yubikey.
If a role's metadata cannot be signed using the inserted yubikey, attempt to sign using keystore files.
If that too is unsuccessful, ask the user if they want to insert the correct ybukey.

Move keystore generation with fixes and yubikey setup to api

Closes #315 
See #317 


taf/tools/roles/__init__.py
n-dusan, line 1: Let's update taf commands in `taf/docs/developers/testing_nodes.md` with the new `roles` module instead of `metadata`

taf/keystore.py
```diff
@@ -94,11 +93,6 @@ def read_private_key_from_keystore(
     if not key_path.is_file():
         raise KeystoreError(f"{str(key_path)} does not exist")
 
-    if password is not None and roles_key_info is not None:
-        passwords = roles_key_info.get("passwords")
-        if passwords is not None and len(passwords):
-            password = passwords[key_num]
-
```
n-dusan, line 101: Why is this not needed?

taf/keys.py
```diff
@@ -102,32 +101,40 @@ def load_signing_keys(
...
+                    keystore, key_name, num_of_signatures, scheme
+                )
+                # load only valid keys
+                if taf_repo.is_valid_metadata_key(role, key, scheme=scheme):
+                    return key
+            except KeystoreError:
+                pass
+        return None
+
+    def _laod_and_append_yubikeys(key_name, role, retry_on_failure):
```
n-dusan, line 124:
```suggestion
    def _load_and_append_yubikeys(key_name, role, retry_on_failure):
```

n-dusan (commented):
🙌!
Do we need to make a PR and update platform calls since we're changing the api from developer_tools to keystore and yubikey?

taf/tools/roles/__init__.py
renatav: Will update

taf/keystore.py
renatav: we don't set passwords, but even if we did, seems like a bad idea to store them in files like that. So I thought it made sense to just remove that

renatav:
@n-dusan `developer_tools` is not used outside of taf cli, so there should be no need to update anything

taf/keystore.py
n-dusan: Makes sense to me

n-dusan (approved):

PR #330, opened 2023-07-27
refact: Remove hosts

### LINKED ISSUE #328: Do not handle hosts and hosts.json
taf/stelae should really be focused on security and serving, not on mapping stelae to hosts

n-dusan:

Closes #328 

Removed hosts and hosts.json logic. It was necessary to remove the host handler and omit the hosts from the pull pipeline schema. Removed hosts from documentation and from testing framework.

To test, run `taf repo update https://github.com/openlawlibrary/open-law`


renatav (commented):
While I agree thwt hosts is not something that TAF necessarily needs to handle (TAF is not meant to be used for hosting purposes, so it being used to determine a host was probably an overkill), 2 questions:
Do we know what was causing the issue leading to the removal of this code? Was it not worth debugging it due to change of plans or was it difficult to debug? Was removal of host handler files not sufficient? If so, then I would definitely want to get to the bottom of it

n-dusan:
I believe the reasoning was that rather than debugging (and getting to the bottom of the issue) we would just remove the host code, while having a reference to this PR so that we could reimplement the handler if it becomes necessary.  So I think the reasoning was that it wasn't worth debugging because our thinking with stelae been to not rely on hosts information.

renatav (approved):
Approving, we can revert the pr if necessary

PR #338, opened 2023-08-16
Add/update/remove dependencies from dependencies.json

### LINKED ISSUE #333: Command add/remove/update dependency in `dependencies.json`

renatav:

Closes #333

Top add/update dependency, run

``` taf dependencies add . cityofsanmateo/law --branch-name master --out-of-band-commit c4dc69c8ed1a9bcd2ff28a8cf791fce492314762 --keystore D:\oll\library\oll-test-repos\openlawlib-test-keystore```

To remove a dependency, run

```
taf dependencies remove auth_path cityofsanmateo/law --keystore D:\oll\library\oll-test-repos\openlawlib-test-keystore
```


n-dusan (changes requested):
Nice work! Here is my feedback about the API.

- What if instead of `taf repo add-dependency, remove-dependency`, etc. we introduce `taf dependency`? Then the API seems much more intuitive: `taf dependency add`, `taf dependency remove`, `update` and so on.
- Should `out-of-band commit` and `branch` need to be specified? It isn't clear how the user would find that information, as they would need to traverse the git commit history. Usually the out-of-band commit is the first commit of the authentication repository. Are there any scenarios in which the librarian cares about the `out-of-band` commit as to manually need to specify it? Can we instead just automate it?
- It does make sense that a commit can be on multiple branches, so it's technically correct that we need add it to dependencies, but is currently only used for auto-detect default branch when running updater. It feels as though we can omit that flag as well, since it's just the default branch of the `repo/law` repository.  We could keep the flag as optional, but instead automate by querying the remote (or local) dependency for the default branch?

If these comments make sense, the API could potentially look something like this:
`taf dependency add <DEPENDENCY> --keystore D:\oll\library\oll-test-repos\openlawlib-test-keystore`

Where  DEPENDENCY is  `cityofsanmateo1/law`. What are your thoughts?

renatav:
@n-dusan I pushed some changes. Having a separate command sounds like a good idea. I named it dependencies because we already have roles and targets. Looking back, not sure if that was a good idea, but decided to keep it consistent with this naming convention (using the plural form).
I also added the possibility to automatically determine branch and commit in the way you described. I had actually initially implemented it like that, but then decided to enforce the specification for a few reasons:
- It might sound technical, but the end users who maintain their repositories need to be aware of this commend. In fact, our initial idea was that someone who wants to clone and host the repositories could contact this maintainer and ask them to confirm value of this commit.
- It might become necessary to set it to a value other than the initial commit. For example, if an attacker compromised the first n commits, it could be set to commit n+1 and such. We want to make sure that this value does not have to represent the first commit on a branch (although there might be parts of our code where we assume that, but that is not fully correct)
- This is a very important piece of information and we do not want it to accidentally be set to an incorrect value due to states of remote and local repositories

So in this implementation, you can provide these values. If you do not, they will be automatically determined, but we'll ask the user to confirm that they want those values to be set

n-dusan (commented):
Tested and LGTM. 

Before merging, do you think we should update documentation for the new CLI as part of this PR? If not, we can just merge. Also, do you think it's worth mentioning in docs that when running `taf dependency add` without extra flags, adding a dependency requires the user to run the updater to download/update the dependency first?

renatav:
That makes sense, I'll update the documentation. Think I'll also update updater docs as a part of this PR following the changes I made in thr other one

n-dusan (approved):

PR #341, opened 2023-08-19
fix: (updater) only load repositories defined in the newest version of repositories.json

renatav:

The problem with loading all repositories that were specified in `repositories.json`, even those that were removed is that a lot of the time the reason for removal is incorrect definition. For example, if there was a typo. At some point, we could implement something that could allow us to differentiate between invalid definitions and repositories that are no longer needed, but for now it makes sense to load only repositories specified in the newest version of `repositories.json`
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Changed
 
+- Updater: only load repositories defined in the newest version of repositories.json ([341])
 - Updater: automatically determined url if local repository exists ([340])
 - Remove hosts and hosts.json ([330])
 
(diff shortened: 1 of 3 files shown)


n-dusan (approved):
Makes sense to me. Since the updater would now load the newest repositories.json, wouldn't it be consistent to also update `repositoriesdb` to only load latest repositories?
However, if my question doesn't make sense, please merge!

renatav:
@n-dusan That makes sense, but I also feel like it's okay to leave `repositoriesdb` flexible. You can load only the latest repositories if that is what you need, you can load different revisions. Loading newest vs loading older is still something that we should discuss in depth. By default, in `repositoriesdb`, there are functions which automatically work with the newest commit

PR #300, opened 2022-12-25
feat: auto-detect default branch

n-dusan:
## Description (e.g. "Related to ...", etc.)

Closes #290

todo:
- [x] get `test_updater.py` tests passing
- [x] add `branch` key under `dependencies.json` so that authentication repositories specify the default branch in json file. then update `repositoriesdb.load_dependencies` with the default branch.

## Code review checklist (for code reviewer to complete)

- [ ] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [ ] Title summarizes what is changing
- [ ] Commit messages are meaningful (see [this][commit messages] for details)
- [ ] Tests have been included and/or updated, as appropriate
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Changelog has been updated, as needed (see [CHANGELOG.md][changelog])

[changelog]: https://github.com/openlawlibrary/taf/blob/master/CHANGELOG.md
[commit messages]: https://chris.beams.io/posts/git-commit/

PR #343, opened 2023-08-23
Remove default branch specification from updater

renatav:

Automatic detection of a repository's default branch was added by #300. Since different repositories can have different default branches (e.g. an authentication repository and one of its target repositories, or an authentication repository and a dependency), specification of one default branch is not sufficient. Automatic detection is a better approach. Tests are passing following the removal of the default branch parameter. I think there are no places in the code where branch name is still hardcoded. It's cleaner to remove the default_branch specification now that it is no longer necessary. If at any point the automatic detection stops working, we can figure out where to store the default_branch name or how to improve the automatic detection. 


n-dusan (changes requested):
Awesome! I don't think I've covered enough of the test cases for default branch auto-detection back in the initial PR, as the only two tests are found [here](https://github.com/openlawlibrary/taf/blob/d72854aef065972364c05c3ed6022103fe06b5da/taf/tests/test_repository.py#L196-L210). Could you please go over the testing framework and add more test cases? I believe what's missing is an already set up git repository with default branch in place. Then assert the default branch once `GitRepository` is instantiated.

n-dusan (approved):

PR #346, opened 2023-08-25
Make it possible to abort signing if keys cannot be loaded, improve cli error handling

### LINKED ISSUE #242: Better update snapshot/timestamp expiration date error message
If the key is incorrect, the error message will say that "snapshot/timestamp" is not a valid delegated role. It should also be more obvious that an error occurred, since it's easy to miss the output when routinely signing using taf cli.

### LINKED ISSUE #334: `taf repo create` `keys-description.json` is not expanding tilda `~`
in `keys-description.json`, having a `"keystore"`: `~/.stelae/keystore/` does not expand to `/home/user/.stelae/keystore/`

renatav:

When loading signing keys, we check if there is a keystore file, then if we can sign using a yubikey before eventually asking the user to manually enter the key (an option that a user might want to use if they want to store the private key in a different way, neither on their filesystem nor on a YubiKey). The problem with this is that in majority of the cases, a user actually wanted to use a keystore file, but provided an incorrect location and they do not want to copy/paste this key. At that point, we had to manually terminate the application. This PR adds a prompt which gives the user an option to say that they do not want to enter this key, in which case an error is raised. This error is caught and the information that the role's metadata could not be signed is printed.

Also added CLI error handling in order to only print error messages and not full stack traces. Only `TAFError` exceptions are caught. In other cases, stack trace will still be printed (meaning that an unexpected error occurred). I think that showing just error messages in case of unexpected errors makes it really difficult to debug. 

Add `prompt_for_keys` flag to all functions that load signing keys. If set to True, we'll ask the user if they want to manually enter the key if not using yubikeys and it cannot be loaded from keystore files. If set to False, the execution of the program will stop if not using Yubikeys and the key is not found inside the provided keystore directory.

Closes #242 
Closes #334

(This was already fixed to some extended in other PRs, but I think that we can fully close it now)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Changed
 
+- Improve CLI error handling ([346])
+- Update signing keys loading. Add a flag for specifying if the user will be asked to manually enter a key ([346])
 - Remove default branch specification from updater ([343])
 - Updater: only load repositories defined in the newest version of repositories.json ([341])
 - Updater: automatically determine url if local repository exists ([340])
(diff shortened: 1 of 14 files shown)


dgreisen:
I don't really see a situation in production where someone would manually type the key in. Instead of a prompt to not enter key, could we add a flag to prompt for keys? Something like --prompt-for-keys? Then if that flag isn't present and key isn't in keystore it immediately fails?

n-dusan (approved):

PR #351, opened 2023-08-29
Add remove target repo fixes

renatav:

- Fix error when running commands in case when keystore is not specified
- When entering custom arguments (in case of commands that include definition of custom data - adding target repos and dependencies), show a nicer error message if they are entered incorrectly and convert "true" and "false" to boolean values
- Fix incorrect `repositories.json` path when adding a new target repo
- Update docstrings
- When removing a delegated path, do not update snapshot and timestamp if the delegated path did not exist and targets metadata was not udpated

### DIFF
--- a/taf/api/roles.py
+++ b/taf/api/roles.py
@@ -127,11 +127,11 @@ def add_role(
         auth_repo.commit(commit_message)
 
 
-@log_on_start(DEBUG, "Adding new paths to role {role:s}", logger=taf_logger)
+@log_on_start(DEBUG, "Adding new paths to role {delegated_role:s}", logger=taf_logger)
 @log_on_end(DEBUG, "Finished adding new paths to role", logger=taf_logger)
 @log_on_error(
     ERROR,
-    "An error occurred while adding new paths to role {role:s}: {e}",
+    "An error occurred while adding new paths to role {delegated_role:s}: {e}",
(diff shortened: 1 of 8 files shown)


taf/tools/cli/__init__.py
```diff
@@ -15,3 +20,34 @@ def wrapper(*args, **kwargs):
...
+
+
+@log_on_error(
+    ERROR,
+    "{e}",
+    logger=taf_logger,
+    on_exceptions=TAFError,
+    reraise=True,
+)
+def process_custom_command_line_args(ctx):
```
n-dusan, line 32: I'm not sure what this function does. Could we add a docstring?

taf/api/roles.py
n-dusan, line 1: So if delegated path does not exist, nothing gets removed (modified)? Might be useful to print a message or log that particular situation

renatav: Good point. We do print a message, but it's not visible in overview. But now that I think about it, I don't like having all of these print statements, no need to not use a logger. Doesn't have to be a part of this PR, but we use both print statements and logger and I'm not sure about it. My thoughts were that we still want to print to console even when logging is disabled, but I'm not sure that that is making much sense.

n-dusan (approved):

PR #353, opened 2023-09-01
Split tests into separate packages

renatav:

One big problem with the testing framework is that setup is slow (test git repositories need to be copied). The majority of these repositories are used for executing updater tests. By splitting tests into separate packages with separate conftest files, we can run a subset of tests without also setting up updater tests, which results in a huge speed improvement. E.g. to run `test_repository`, navigate into package of the same name, run the tests and its execution time will be around 2-3 second now, while previously it was necessary to wait much more than that because updater repositories were getting prepared - completely unnecessarily.

Also bumped pytest version


n-dusan (approved):
lgtm. Are tests 40% faster now? Not too shabby!

renatav:
> lgtm. Are tests 40% faster now? Not too shabby!

Not sure about the performance when running all tests together, but it should be way faster to run an individual tests (apart from updater tests)

PR #360, opened 2023-09-26
Set up mypy and fix the issues it reported

### LINKED ISSUE #287: Add mypy for static type checking
Setup _mypy_ for static type checking

http://mypy-lang.org/
https://mypy.readthedocs.io/en/stable/

renatav:

- Added mypy pre-commit check. Use version 1.4.1, which is the last version that supports Python 3.7
- Addressed a variety of issues that mypy reported, mostly:
  - variable of one type being assigned a value of a different type, like when converting a string to a Path instance
  - error handling when a variable can be `None`
  - incorrect typing clues that were previously added
  - rework of roles description models, remove `DelegatedRole`, `TargetsRole` can now contain a list of delegated `TargetsRole` objects
- Some generation of keys fixes, in cases when the keystore should have a password or when it should be printed to console
- Added additional type hints

Closes #287 

Did no test cli, it's possible that this will introduce some regressions, but since the next big chunk of work will add api tests, all issues, if there are any, will be addressed then

### DIFF
--- a/taf/constants.py
+++ b/taf/constants.py
@@ -3,30 +3,23 @@
 import datetime
 from typing import List, Optional
 
+import attrs
+
 
 DEFAULT_RSA_SIGNATURE_SCHEME = "rsa-pkcs1v15-sha256"
 
 CAPSTONE = "capstone"
 
 
+@attrs.define
 class RoleSetupParams:
-    def __init__(
-        self,
-        number: int = 1,
-        threshold: int = 1,
-        yubikey: bool = False,
-        scheme: str = DEFAULT_RSA_SIGNATURE_SCHEME,
-        length: int = 3072,
-        passwords: Optional[List[str]] = None,
-        terminating: bool = True,
-    ):
-        self.number = number
-        self.threshold = threshold
-        self.yubikey = yubikey
-        self.scheme = scheme
-        self.length = length
-        self.passwords = passwords
-        self.terminating = terminating
+    number: int = attrs.field(default=1)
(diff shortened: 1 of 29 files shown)


taf/constants.py
```diff
@@ -1,21 +1,37 @@
...
-    "number": 1,
-    "threshold": 1,
-    "yubikey": False,
-    "scheme": DEFAULT_RSA_SIGNATURE_SCHEME,
-    "length": 3072,
-    "passwords": None,
-    "terminating": True,
-}
+
+class RoleSetupParams:
```
n-dusan, line 12: Could we use attrs instead of a regular class?

taf/git.py
```diff
@@ -401,6 +419,9 @@ def branch_exists(
         a remote branch exists.
         """
         repo = self.pygit_repo
+        if repo is None:
+            # TODO default to _git
+            return False
```
n-dusan, line 424: Do we want this code to raise `GitError`? It looks like we're raising errors on `if repo is None`, but not here?

n-dusan (changes requested):
Two minor comments. This looks really good!

taf/git.py
renatav: I think the only reason is that this was the first such mypy error that I tried addressing and I initially thought that pygit2 instantiaon errors would get raised somewhere, then realized that that would not be the case and that we would return incorrect values from these functions.

n-dusan (approved):

PR #370, opened 2023-11-09
Fix loading of keys and  create repo when old yubikey flag is used

renatav:

When loading signing keys, we want to first try reading from keystore files, then from the inserted Yubikey, before asking the user to do anything. The problem was that the check if key was already read from a YubiKey was not properly implemented, so the same key was getting read twice. Also tried to improve the overall user experience when loading keys.
Found issues with creating a repository in case when old `yubikey` flag was used to specify that a role's metadata files should be signed using YubiKeys.

### DIFF
--- a/taf/models/types.py
+++ b/taf/models/types.py
@@ -57,7 +57,7 @@ class Role:
     @property
     def yubikey_ids(self):
         if not self.is_yubikey:
-            return []
+            return None
         if self.yubikeys:
             return self.yubikeys
         if self.number == 1:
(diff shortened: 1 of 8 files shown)


taf/models/types.py
```diff
@@ -54,6 +54,16 @@ def is_yubikey(self):
             self.yubikey is True or (self.yubikeys is not None and len(self.yubikeys))
         )
 
+    @property
+    def yubikey_ids(self):
+        if not self.is_yubikey:
+            return []
```
n-dusan, line 60: I believe this should return None, since in `taf/keys.py`  L307, L308 there is a check for None

n-dusan (approved):

PR #376, opened 2024-01-10
Fix imports, do not require installation of ykman in order to run the updater

renatav:

- Made it possible to run the updater without installing yubikey manager
- Fixed a `GitError` impor
- Fixed a failing tests following update of the imports. The behavior of this test is now the same as before the updater rework

### DIFF
--- a/taf/api/utils/_roles.py
+++ b/taf/api/utils/_roles.py
@@ -5,7 +5,6 @@ from logdecorator import log_on_end, log_on_start
 from tuf.repository_tool import Repository as TUFRepository, Targets
 from taf.exceptions import TAFError
 from taf.models.types import RolesIterator
-from taf.yubikey import get_key_serial_by_id
 from tuf.repository_tool import Metadata
 from taf.keys import get_key_name
 from taf.auth_repo import AuthenticationRepository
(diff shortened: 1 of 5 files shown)


taf/api/utils/_roles.py
```diff
@@ -18,6 +17,11 @@
 from taf.models.types import Role
 from taf.log import taf_logger
 
+try:
+    from taf.yubikey import get_key_serial_by_id
+except ImportError:
+    pass
+
```
n-dusan, line 24: question: Won't we need taf yubikey to work with setting up roles?

renatav: We won't need them if we only want to use keystore files. We are not able to run any commands without yubikey manager unless this import is moved directly to the function that is using it or surrounded with try/except. If you do not have yubikey manager installed, a warning message saying that you are not able to use yubikeys will be shown when you run any of taf commands: "WARNING: yubikey-manager dependency not installed. You will not be able to use YubiKeys." I have not tested what it the error will actually look like if you try setting up roles with no yubikey manager

n-dusan (approved):

PR #380, opened 2024-01-17
Separate clone and update commands and read repository name from info.json when cloning

### LINKED ISSUE #378: Separate clone and update

renatav:

Separated clone and update commands: This separation mirrors Git, where cloning and pulling are separate operations.

The clone command is used for the initial setup of repositories.
If the `--path` option is not provided during cloning, the updater automatically reads from `protected/info.json` to determine the repository's location. `library-dir`, if nor provided, is set to the current directory and the namespace and name read from `info.json` are appended to this path to determine authentication repository's location.

The update command is designed to pull changes into an already cloned repository.
When updating, the URL of the repository is automatically determined and is not expected to be specified.
If the --path option is not provided, the command defaults to using the current directory as the repository's location.

Updated docs.

So, to clone, run for example:

```git clone https://github.com/oll-test-repos/sanipueblo-law```

This repository contains `protected/info.json`, so it is not necessary to use the `--path` option

```git clone https://github.com/sanipueblo/law --path sanipueblo/law```

This repository does not contain `info.json`, so path needs to be specified.

Update was not modified, but it can only be run if a repository already exists. An error will be raised if that is not the case.

Closes  #378

### DIFF
--- a/taf/tools/repo/__init__.py
+++ b/taf/tools/repo/__init__.py
@@ -252,7 +252,7 @@ def attach_to_group(group):
 
     @repo.command()
     @click.option("--path", default=".", help="Authentication repository's location. If not specified, set to the current directory")
-    @click.option("--clients-library-dir", default=None, help="Directory where target repositories and, "
+    @click.option("--library-dir", default=None, help="Directory where target repositories and, "
                   "optionally, authentication repository are located. If omitted it is "
                   "calculated based on authentication repository's path. "
                   "Authentication repo is presumed to be at library-dir/namespace/auth-repo-name")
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: taf/tools/repo/__init__.py [hunk 1]
```diff
@@ -133,30 +143,90 @@ def update(path, url, clients_library_dir, from_fs, expected_repo_type,
...
-        if path is None and clients_library_dir is None:
+        if path is None and library_dir is None:
             raise click.UsageError('Must specify either authentication repository path or library directory!')
 
         if profile:
-            import cProfile
-            import atexit
+            start_profiling()
+
+        config = RepositoryConfig(
```
n-dusan, line 150: Can we get into a situation where we unintentionally misuse the new RepositoryConfig when we forget to specify some field that's expected in a function? Do we need to add validation to all functions that accept RepositoryConfig object that check if all necessary fields are specified (that that function expects to use), before continuing with other code?

n-dusan (approved):
One question about the new attrs config class. If you don't think it's a problem, please merge!

taf/tools/repo/__init__.py
renatav: The check is done in `updater.py`. For example:
```
   if config.url is None:
        raise UpdateFailedError("URL has to be specified when cloning repositories")
```
Default values are also set here. So I think that it should be fine

PR #384, opened 2024-01-25
Improve error messages and minor fixes

renatav:

- Fix `_update_named_repository` call
- Remove an unused function
- Fix update of referenced authentication repositories in case when the parent repository exists and the child does not (caused by the recent separation of clone and update functionalities)
- Improve error reporting

### DIFF
--- a/taf/updater/lifecycle_handlers.py
+++ b/taf/updater/lifecycle_handlers.py
@@ -59,10 +59,15 @@ def _get_script_path(lifecycle_stage, event):
 def get_config(library_root, config_name=CONFIG_NAME):
     global config_db
     if library_root not in config_db:
+        config_path = Path(library_root, CONFIG_NAME)
         try:
-            config = json.loads(Path(library_root, CONFIG_NAME).read_text())
+            config = json.loads(Path(config_path).read_text())
         except Exception:
             config = {}
+            taf_logger.warning(
+                "Config file {} not found. Scripts might not be executed successfully",
+                config_path,
+            )
         config_db[library_root] = config or {}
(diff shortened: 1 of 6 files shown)


taf/updater/lifecycle_handlers.py
```diff
@@ -242,6 +247,11 @@ def prepare_data_repo(
     error,
     targets_data,
 ):
+    conf_data = get_config(auth_repo.library_dir)
+    if not conf_data:
+        taf_logger.warning(
+            "Warning: No config data. Scripts might not be executed successfully"
```
n-dusan, line 253: Does `taf_logger.warning` already prefix "Warning: "? If it does, no need to include `Warning:`. If it doesn't, then need to include "Warning: ` to the other logging message above

n-dusan (commented):
Thanks for catching this bug. I do remember us testing some previous updater pr a while back and a similar bug crept up, but we only caught it during manual testing. My thoughts: it doesn't seem that complicated to add a simple test-case of a git repository with `dependencies.json` to two other child repositories. Then the test-case runs the updater and confirms everything is updated. That test-case would hopefully catch any future bugs with top level authentication repository updater. On the other hand, it's unpleasant/tedious to add fixtures manually and we would need to manually set that up. What are your thoughts, is adding that test-case more useful than not having it right now?

taf/updater/lifecycle_handlers.py
renatav: It does not print Warning when printing to console. There was another logger call which printed WARNING, so I updated all to match that

n-dusan (approved):

PR #386, opened 2024-02-01
Add test for updating a repositories which references other authentication repositories

renatav:

This test programmatically sets up repositories instead of adding a previously created repository with .git folder renamed to git
Also fixed a minor bug where update status was incorrectly being set in case when a repository with only one commit is cloned

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,15 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Added
 
+- Add a test for updating a repositories which references other authentication repositories. Test repositories are set up programmatically ([386])
+
 ### Changed
 
 ### Fixed
+
+- Fix a minor bug where update status was incorrectly being set in case when a repository with only one commit is cloned
+
+[386]: https://github.com/openlawlibrary/taf/pull/386
 
 ## [0.29.0] - 01/24/2024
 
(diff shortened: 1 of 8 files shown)


taf/tests/test_updater/test_repo_update/conftest.py
```diff
@@ -1,12 +1,83 @@
...
+        auth_repo = GitRepository(path=auth_repo_path)
+        auth_repo.init_repo()
+        repositories_path_template = (TEST_INIT_DATA_PATH / "repositories.json").read_text()
+        repositories_json_str = re.sub(r'\{namespace\}', namespace, repositories_path_template)
+        repositories_json = json.loads(repositories_json_str)
+
+        # Write repositories JSON to file
+        targets_dir = auth_repo_path / "targets"
+        targets_dir.mkdir(exist_ok=True)
+        repositories_json_path = targets_dir / "repositories.json"
```
n-dusan: Since we're using static variables, might make sense to reference a global variable called `REPOSITORIES_JSON` or similar instead. Same for mirrors and dependencies

CHANGELOG.md
```diff
@@ -9,10 +9,16 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Added
 
+- Add a test for updating a repositories which references other authentication repositories. Test repositories are set up programmatically ([386])
+
 ### Changed
 
 ### Fixed
 
+- Fix a minor bug where update status was incorrectly being set in case when a repository with only one commit is cloned
```
n-dusan, line 18: Need to reference a pr number here

n-dusan (approved):

PR #389, opened 2024-02-15
Check if metadata files at revision match those downloaded by TUF updater

### LINKED ISSUE #385: Update does not fail when `snapshot` is not valid
When `timestamp` is invalid, update fails. If `snapshot` is updated in the same invalid way, it does not.

renatav:

TUF updater does not always check the validity of all metadata files
if timestamp is not updated, the updater will determine that a new version
of the snapshot file does not need to be downloaded and it will not be validated
during the update process, the metadata files that TUF updater downloads is stored
in a separate folder within the temp directory
 For each commit, check if the metadata files inside that directory are the same
 as the ones in the auth repository's metadata folder at that revision

Closes #385 

### DIFF
--- a/taf/updater/updater_pipeline.py
+++ b/taf/updater/updater_pipeline.py
@@ -1227,9 +1227,16 @@ def _run_tuf_updater(git_updater, auth_repo_name):
                     git_updater.metadata_dir, metadata_file_name
                 )
                 if not current_tuf_metadata_file.is_file():
-                    raise UpdateFailedError(
-                        f"Invalid metadata file {metadata_file_name}"
-                    )
+                    # this validation causes an issue with one of the first
+                    # commits of our production repositories and it should
+                    # not be enabled until we specify a later commit of those
+                    # repositories as the initial valid ones
+                    # this error happens when a metadata file is added, but
+                    # snapshot is not updated
+                    # raise UpdateFailedError(
+                    #     f"Invalid metadata file {metadata_file_name}"
+                    # )
+                    continue
                 metadata_content = git_updater.get_current_metadata_data(
                     metadata_file_name
                 )
--- a/taf/updater/updater_pipeline.py
+++ b/taf/updater/updater_pipeline.py
@@ -1184,12 +1184,13 @@ def _run_tuf_updater(git_fetcher, auth_repo_name):
             taf_logger.error(f"Failed to instantiate TUF Updater due to error: {e}")
             raise e
 
-
     last_validated_commit = None
     try:
         while not git_fetcher.update_done():
             updater = _init_updater()
-            current_commit = _update_tuf_current_revision(git_fetcher, updater, auth_repo_name)
+            current_commit = _update_tuf_current_revision(
+                git_fetcher, updater, auth_repo_name
+            )
             if current_commit is not None:
                 last_validated_commit = current_commit
     except UpdateFailedError as e:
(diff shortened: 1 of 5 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -1205,6 +1207,45 @@ def _update_tuf_current_revision():
...
                     current_commit,
                 )
+
+            # TUF updater does not always check the validity of all metadata files
+            # if timestamp is not updated, the updater will determine that a new version
+            # of the snapshot file does not need to be downloaded and it will not be validated
+            # during the update process, the metadata files that TUF updater downloads is stored
+            # in a separate folder within the temp directory
+            # For each commit, check if the metadata files inside that directory are the same
+            # as the ones in the auth repository's metadata folder at that revision
```
n-dusan, line 1217: This explanation looks like a good candidate to be a docstring. Could we move the code to a function?

taf/updater/updater_pipeline.py
```diff
@@ -1205,6 +1207,45 @@ def _update_tuf_current_revision():
...
+            pattern = r"\d+\.[^\.\s]+\.\w+"
```
n-dusan, line 1218: Might want to add explanation as to what this pattern is searching for

taf/updater/updater_pipeline.py
```diff
@@ -1205,6 +1207,45 @@ def _update_tuf_current_revision():
...
+                if not current_tuf_metadata_file.is_file():
+                    # this validation causes an issue with one of the first
+                    # commits of our production repositories and it should
+                    # not be enabled until we specify a later commit of those
+                    # repositories as the initial valid ones
+                    # this error happens when a metadata file is added, but
+                    # snapshot is not updated
+                    # raise UpdateFailedError(
+                    #     f"Invalid metadata file {metadata_file_name}"
+                    # )
```
n-dusan, line 1238: Could we turn this into an issue?

taf/updater/updater_pipeline.py
```diff
@@ -1205,6 +1207,45 @@ def _update_tuf_current_revision():
...
+
+            # TUF updater does not always check the validity of all metadata files
+            # if timestamp is not updated, the updater will determine that a new version
+            # of the snapshot file does not need to be downloaded and it will not be validated
+            # during the update process, the metadata files that TUF updater downloads is stored
+            # in a separate folder within the temp directory
+            # For each commit, check if the metadata files inside that directory are the same
+            # as the ones in the auth repository's metadata folder at that revision
+            pattern = r"\d+\.[^\.\s]+\.\w+"
+            for metadata_file_name in git_updater.get_current_metadata():
```
n-dusan, line 1219: Since we're reading all metadata files from disk for each commit, it's worth timing the updater to see how long this new validation takes, compared to the implementation we have on master.

n-dusan (changes requested):

taf/updater/updater_pipeline.py
renatav: Ran SMC clone a couoek of times. The first was was slightly slower than on master, the second faster than on master, so I don't think that this impacts the performance. Other operations are much slower

n-dusan (approved):

taf/updater/updater_pipeline.py
n-dusan: Awesome

PR #404, opened 2024-03-19
Improve merge commits

### LINKED ISSUE #395: Do not merge target repos when the event is unchanged
Currently even when there are no changes to update, the updater merges some commits to the target repositories., which shouldn't happen.

First check and see if there are any changes for target repos, then try merging

renatav:

- Fix logic for determining if a commits should be merged
- Remove checkout option since we were always checking out the branch. After cloning, if a branch is not checked out, the index will end up being dirty, showing that the repository's content was removed. We could potentially avoid this checkout by checking if the repository was cloned and then make sure that nothing else fails without it. This can be done later.
- Only show "Merging commits" message  if commits are actually getting merged, and not always before and after that function is called

Closes #395

### DIFF
--- a/taf/git.py
+++ b/taf/git.py
@@ -338,6 +338,8 @@ class GitRepository:
             )
         if branch:
             branch_obj = repo.branches.get(branch)
+            if branch_obj is None:
+                return []
             latest_commit_id = branch_obj.target
         else:
             latest_commit_id = repo[repo.head.target].id
--- a/taf/git.py
+++ b/taf/git.py
@@ -752,6 +752,36 @@ class GitRepository:
             reraise_error=True,
         )
 
+    def check_if_unpushed_commits(self, branch_name):
+        repo = self.pygit_repo
+        if repo is None:
+            raise GitError(
+                self,
+                message="pygit repository could not be instantiated.",
+            )
+
+        # Ensure the branch exists locally.
+        local_branch_ref = f"refs/heads/{branch_name}"
+        if local_branch_ref not in repo.references:
+            print(f"Branch '{branch_name}' does not exist.")
+            return unpushed_commits
+
+        # Ensure the branch has an upstream.
+        try:
+            upstream_branch_ref = repo.lookup_branch(branch_name).upstream.name
+        except ValueError:
+            print(f"Branch '{branch_name}' does not have an upstream set.")
+            return unpushed_commits
+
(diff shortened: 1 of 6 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -1329,19 +1315,32 @@ def _merge_commit(
...
-        repository.checkout_branch(branch)
+
+    elif len(commits_since_to_merge) and force_revert:
+        taf_logger.info(
+            "{} Reverting branch {} to {}",
+            repository.name,
+            branch,
+            format_commit(commit_to_merge),
+        )
+        repository.reset_to_commit(commit_to_merge, hard=True)
```
n-dusan, line 1339: Is the reason why we don't merge the `commit_to_merge` n this condition because merging a newer commit will lead to fast forward? If that's the case, might want to write that down somewhere in here, since it seems confusing as to why we're resetting to commit?

n-dusan (approved):
LGTM. Have a question

taf/git.py
```diff
@@ -752,6 +752,36 @@ def checkout_commit(self, commit: str) -> None:
             reraise_error=True,
         )
 
+    def check_if_unpushed_commits(self, branch_name):
```
n-dusan, line 755: `check_if` doesn't sound like a boolean method. What about `is_branch_with_unpushed_commits`?

taf/updater/updater_pipeline.py
```diff
@@ -674,6 +681,17 @@ def get_targets_data_from_auth_repo(self):
         self.state.target_branches_data_from_auth_repo = repo_branches
         return UpdateStatus.SUCCESS
 
+    def check_if_local_repositories_contain_unpushed_commits(self):
```
n-dusan, line 684: `error_if_local_repositories_with_unpushed_commits`?

taf/git.py
```diff
@@ -1367,6 +1367,14 @@ def reset_to_commit(self, commit: str, hard: Optional[bool] = False) -> None:
         flag = "--hard" if hard else "--soft"
         self._git(f"reset {flag} {commit}")
 
+    def reset_remote_tracking_branch(self, branch_name) -> None:
+        """
+        Set top commit of origing/branch to the top comit of the local branch
```
n-dusan, line 1372: `origin/branch`

n-dusan (approved):

PR #412, opened 2024-03-31
Clone target repositories to temp

### LINKED ISSUE #408: Clone/update target repositories inside temp directory before updating content on disk
If the repository does not exist, clone to temp first, like auth repo. If it does, create in temp/link to the version already on disk and then fetch changes. Pull changes after validation.

renatav:

Do not clone target repositories directly to their permanent location. Clone to temp first, just like when working with authentication repositories. If a repository is already on disk, copy that to temp, update temp, then fetch changes from temp if valid.

Note: This PR does not address the issue of not cloning when running validation. That can be done in a separate PR

Closes #408 

### DIFF
--- a/taf/git.py
+++ b/taf/git.py
@@ -644,7 +644,7 @@ class GitRepository:
             self.default_branch = self._determine_default_branch()
 
     def clone_from_disk(
-        self, local_path: Path, remote_url: str, is_bare: bool = False
+        self, local_path: Path, remote_url: Optional[str] = None, is_bare: bool = False
     ) -> None:
         self.path.mkdir(parents=True, exist_ok=True)
         pygit2.clone_repository(local_path, self.path, bare=is_bare)
@@ -656,11 +656,12 @@ class GitRepository:
                 "Cloning from disk could not be completed. pygit repo could not be instantiated"
             )
         self.remove_remote("origin")
-        self.add_remote("origin", remote_url)
-        self.fetch()
-        if repo is not None:
-            for branch in repo.branches.local:
-                self.set_upstream(str(branch))
+        if remote_url is not None:
+            self.add_remote("origin", remote_url)
+            self.fetch()
+            if repo is not None:
+                for branch in repo.branches.local:
+                    self.set_upstream(str(branch))
 
     def clone_or_pull(
         self,
@@ -1036,8 +1037,9 @@ class GitRepository:
         sort_key_func: Optional[Callable[[str], bool]] = None,
     ) -> Optional[str]:
         """
-        Fetches changes from a local repository on disk
-        and temporarily adds it as a remote to the current repository.
+        Fetches changes from a local repository on disk.
+        Temporarily adds it as a remote to the current repository
+        and removes that remote after the operation is completed.
         """
 
         branch_tips = {}
--- a/taf/git.py
+++ b/taf/git.py
@@ -4,6 +4,7 @@ import json
 import itertools
 import os
 import re
+import uuid
 import pygit2
 import subprocess
 import logging
@@ -642,11 +643,17 @@ class GitRepository:
         if self.default_branch is None:
             self.default_branch = self._determine_default_branch()
 
-    def clone_from_disk(self, local_path: Path):
+    def clone_from_disk(self, local_path: Path, remote_url: str):
         urls_backup = list(self.urls)
         self.urls = [str(local_path.resolve())]
         self.clone()
         self.urls = urls_backup
+        self.remove_remote("origin")
+        self.add_remote("origin", remote_url)
+        self.fetch()
+        repo = self.pygit_repo
+        for branch in repo.branches.local:
+            self.set_upstream(str(branch))
 
     def clone_or_pull(
         self,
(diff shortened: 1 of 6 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -557,18 +578,25 @@ def check_if_local_target_repositories_clean(self):
...
+            self.state.repos_not_on_disk = []
+            for temp_repo in self.state.temp_target_repositories.values():
+                users_repo = self.state.users_target_repositories[temp_repo.name]
+                is_git_repository = users_repo.is_git_repository_root
+                if is_git_repository:
+                    temp_repo.clone_from_disk(
+                        users_repo.path,
+                        users_repo.get_remote_url(),
+                        is_bare_repository=True,
+                    )
```
n-dusan, line 595: question: Will this do a complete clone of the target repositories to temp dir? I can imagine a set of target repositories having gigabytes of data, (maybe of tens of gigabytes, if cloning a top-level stele). Is this approach of a full local clone going to scale if the root authentication repository references over 10-20 other authentication repos?

n-dusan (changes requested):
Nice work! I left some questions/suggestions below

taf/updater/updater_pipeline.py
```diff
@@ -989,6 +1024,46 @@ def validate_and_set_additional_commits_of_target_repositories(self):
...
+            else:
+                users_target_repo.fetch_from_disk(temp_target_repo.path)
+
+        if self.state.update_status == UpdateStatus.FAILED:
+            return self.state.update_status
+        try:
+            for name in self.state.repos_not_on_disk:
+                _clone_or_copy_from_disk(name, True)
+            for name in self.state.repos_on_disk:
+                _clone_or_copy_from_disk(name, False)
```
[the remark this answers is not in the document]
renatav: I was just trying to avoid setting users_target_repo and temp_target_repo twice, but probably not an issue. I will remove the inner function

taf/updater/updater_pipeline.py
```diff
@@ -557,18 +578,25 @@ def check_if_local_target_repositories_clean(self):
             self.state.event = Event.FAILED
             return UpdateStatus.FAILED
 
-    @log_on_start(
-        INFO, "Cloning target repositories which are not on disk...", logger=taf_logger
-    )
+    @log_on_start(DEBUG, "Cloning target repositories to temp...", logger=taf_logger)
     @log_on_end(INFO, "Finished cloning target repositories", logger=taf_logger)
-    def clone_target_repositories_if_not_on_disk(self):
+    def clone_target_repositories_to_temp(self):
```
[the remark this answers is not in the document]
renatav: At the moment, we neither have tests for individual pipeline steps, nor git. We can create issues to cover both with unit tests. I added a basic test for cloning from local, which can be a starting point.

taf/updater/updater_pipeline.py
renatav: That did cross my mind. I added except `KeyboardInterrupt`. Need to figure out how to integrate with codify

taf/updater/updater_pipeline.py
```diff
@@ -506,13 +522,28 @@ def load_target_repositories(self):
...
                     self.state.auth_commits_since_last_validated[-1::],
                 )
             )
-            return UpdateStatus.SUCCESS
+            if self.only_validate:
+                self.state.temp_target_repositories = (
+                    self.state.users_target_repositories
+                )
+            else:
+                self.state.temp_dir = TemporaryDirectory()
```
n-dusan, line 536: I don't think we should create a temporary directory at default disk location, right? Because it's often the case that the clone crosses filesystem boundaries. This isn't a problem for the authentication repo, which is often much smaller. But it can be a significant impact to have write operations from D:// to C:// disk. So the clone is expected to be much slower, IIUC. We can solve by creating a temporary directory at the same disk/space that the archive is located at

taf/updater/updater_pipeline.py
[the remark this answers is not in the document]
n-dusan: Thanks. We should really have those tests

taf/git.py
```diff
@@ -642,6 +643,26 @@ def clone(
         if self.default_branch is None:
             self.default_branch = self._determine_default_branch()
 
+    def clone_from_disk(
+        self, local_path: Path, remote_url: Optional[str] = None, is_bare: bool = False
+    ) -> None:
```
n-dusan, line 648: I ran taf repo update on my local repositories and new local clone feels really slow (I didn't run a profiler). The problem is that we want updater performance to get faster 😞.

We should discuss/investigate if there are any approaches which we haven't considered here, since we want updater performance. Since the local disk clone is now slower, are symlinks disabled by default for git? I thought the local disk clone was supposed be instantaneous? 
Does this do a full copy of the directory contents? If so, I'd expect this to be a problem for law-html, since the repo size can get gigantic (>3 GB).  We'll also have a problem on the dev server, where the disk volume is 30GB, so I don't think we can reasonably expect to have full clones for 10 authentication repositories to work, right?

n-dusan (commented):
I like the changes and I think we can merge when we address the two issues I've found and left inline. We can also pair and figure out possible solutions together if that would be helpful

taf/git.py
renatav: We do want the performance to be improved, but this feature is aimed at addressing security concerns. We used to clone target repositories directly to their final location, just without checkout. Fetch also pulls down new changes, but will not merge them unless validation is successful. However, if it is not successful, these fetched chances will not get removed. This PR will first clone repositories to temp as bare repositories and validate them. If the validation is successful, they are cloned from that temp directory. These temp repositories are removed in any case. When running update, the repositories are clone from user's repository to temp, changes are fetched, and if the validation succeeds, user's repositories are updated by fetching from disk. So, we have new operations on top of everything else that we were doing, so this can, at best, be slightly slower than it was. This feature was not developed as a part of performance improvement efforts. We should figure out how much slower this is and if the performance can stay roughly the same as it was.

taf/updater/updater_pipeline.py
renatav: It could help, I pushed these changes.

n-dusan (approved):
Went over the code and works as expected. Let's work on performance in a separate pr.

PR #415, opened 2024-04-04
fix is_branch_with_unpushed_commits

### LINKED ISSUE #414: After updater revert a branch to the last published commit, the following update fails due to there being unpushed commits
The check if commits were pushed or not needs to be improved

renatav:

This bug was introduced by unreleased changes, which made it impossible to run update after cloning in cases where
there are authenticated commits that do not get merged. The check needs to be more complex than just checking if
top commits are the same.

Fix #414

### DIFF
--- a/taf/git.py
+++ b/taf/git.py
@@ -644,7 +644,11 @@ class GitRepository:
             self.default_branch = self._determine_default_branch()
 
     def clone_from_disk(
-        self, local_path: Path, remote_url: Optional[str] = None, is_bare: bool = False
+        self,
+        local_path: Path,
+        remote_url: Optional[str] = None,
+        is_bare: bool = False,
+        keep_remote=False,
     ) -> None:
         self.path.mkdir(parents=True, exist_ok=True)
         pygit2.clone_repository(local_path, self.path, bare=is_bare)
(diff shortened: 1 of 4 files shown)


taf/git.py
```diff
@@ -781,26 +781,29 @@ def is_branch_with_unpushed_commits(self, branch_name):
...
-        try:
-            upstream_branch_ref = repo.lookup_branch(branch_name).upstream.name
-        except ValueError:
-            # not upstream yet
+        upstream_full_name = local_branch.upstream_name
+        if not upstream_full_name:
+            # no upstream branch - not pushed
             return True
+        parts = upstream_full_name.split("/")
+        upstream_name = "/".join(parts[2:])
```
n-dusan, line 791: Why does `upstream_name` start with '/'?

n-dusan (commented):
LGTM. How hard would it be to unit test this with our new repo creation testing framework?

taf/git.py
renatav: It's `refs/remotes/origin/branch_name`, so we remove the first 2 parts

renatav:
@n-dusan I added a test, and it's passing locally, but failing on CI. I don't know why. Would require debugging. It does not appear to be failing due to an error with this updated function

n-dusan:
Looks like repo can't find the local git branch name?

```python
local_branch = repo.branches.get(branch_name)
        if local_branch is None:
            print(f"Branch '{branch_name}' not found in the repository.")
upstream_full_name = local_branch.upstream_name
```
  
E       AttributeError: 'NoneType' object has no attribute 'upstream_name'

renatav:
@n-dusan So it turned out that `.default_branch` was returning `renatav/fix-clean-checks` and not `master`. That caused the tests to fail. I have never experienced any problems with `default_branch` until now. The tests are passing now.

n-dusan (approved):
That's very odd, we'll figure out default branch on ci issue in another pr.

PR #416, opened 2024-04-09
Check if repositories are clean before running the updater

### LINKED ISSUE #390: Move `checking if all existing target repositories are clean` to earlier step during `taf repo update`
- Make the updater fail because target repos aren't clean earlier in the process. Currently the updater runs the full update/validation process, then fails because targets have uncommitted changes. 
- Updater fails if untracked files exist in the repository, like .vscode settings (which aren't in gitignore). Think through if the check for clean repository can keep untracked changes and still update

Error stack trace:
```
WARNING: Could not validate authentication repository at revision 2bb34e240fde727c2d8da66f47fa29eb42a2c636 due to error: Final root.json is expired
Successfully validated authentication repository cityofsanmateo/law
Validating out of band commit and update type
Checking if local auth repo is clean...
Cloning or updating user's authentication repository...
Checking if all existing target repositories are clean...
An error occurred while updating repository cityofsanmateo/law while running step check_if_local_target_repositories_clean: Repository cityofsanmateo/law-xml-codified has uncommitted changes. Commit or revert the changes and run the command again.
Called LifecycleStage.UPDATE handler. Event: failed
WARNING: Config file D:\library\partners\config.json not found. Scripts might not be executed successfully
Update of cityofsanmateo/law failed due to error: Repository cityofsanmateo/law-xml-codified has uncommitted changes. Commit or revert the changes and run the command again.
```

renatav:

The idea is for the update process to fail early if the auth repo or one of the existing target repositories is not clean (index is dirty or if there are unpushed commits). After updating the authentication repository, we can check other branches.

Closes #390

### DIFF
--- a/taf/updater/updater_pipeline.py
+++ b/taf/updater/updater_pipeline.py
@@ -204,20 +204,20 @@ class AuthenticationRepositoryUpdatePipeline(Pipeline):
 
         super().__init__(
             steps=[
+                (self.set_existing_repositories, RunMode.UPDATE),
+                (self.check_if_local_repositories_clean, RunMode.UPDATE),
                 (self.clone_remote_and_run_tuf_updater, RunMode.ALL),
                 (self.validate_out_of_band_and_update_type, RunMode.ALL),
-                (self.check_if_local_auth_repo_clean, RunMode.UPDATE),
                 (self.clone_or_fetch_users_auth_repo, RunMode.UPDATE),
                 (self.load_target_repositories, RunMode.ALL),
                 (self.check_if_repositories_on_disk, RunMode.LOCAL_VALIDATION),
-                (self.check_if_local_target_repositories_clean, RunMode.UPDATE),
+                (self.clone_target_repositories_to_temp, RunMode.UPDATE),
+                (self.determine_start_commits, RunMode.ALL),
+                (self.get_targets_data_from_auth_repo, RunMode.ALL),
                 (
                     self.check_if_local_repositories_contain_unpushed_commits,
                     RunMode.UPDATE,
                 ),
-                (self.clone_target_repositories_to_temp, RunMode.UPDATE),
-                (self.determine_start_commits, RunMode.ALL),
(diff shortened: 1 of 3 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -204,20 +204,20 @@ def __init__(
...
-                    self.check_if_local_repositories_contain_unpushed_commits,
-                    RunMode.UPDATE,
-                ),
                 (self.clone_target_repositories_to_temp, RunMode.UPDATE),
                 (self.determine_start_commits, RunMode.ALL),
                 (self.get_targets_data_from_auth_repo, RunMode.ALL),
+                (
+                    self.check_if_local_target_repositories_clean,
+                    RunMode.UPDATE,
+                ),
```
n-dusan, line 220: This method checks if local target repositories contain unpushed commits? If so, why don't we instead early exit updater if local target contains unpushed commits? Then `self.clone_target_repositories_to_temp` step comes after we've already checked that target repos can be updated. This way we avoid the extra work of cloning target repositories to temp (if the target repo isnt' clean). Am I missing something?

renatav: In `check_if_local_repositories_clean`, which is run before we try updating the repositories, we check if repositories which are currently on disk are clean and if the most recent published branch (according to the users' authentication repository, before update) contains unpushed commits. This step is run after we fetch the changes and collect the newest data (including a list of branches per repositories which were published). It needs to be run after that information is known.

n-dusan: I was thinking we would use git ls-remote. But the command doesn't look like it can list all commits on remote branch :-(.

n-dusan (approved):
LGTM (I did not test locally).

taf/updater/updater_pipeline.py
renatav: We have a `get_last_remote_commit` method which uses `ls-remote` to find the last remote commit. We could speed things up by using that to check if there are any new commits that should be pulled down. Can be done in a separate PR

PR #422, opened 2024-05-20
feat:new function taf targets status that prints state of whole library

ronantakizawa:

New function taf targets status that prints the state of whole library


taf/api/targets.py
```diff
@@ -662,3 +662,83 @@ def _update_target_repos(
...
+    repositories_data = auth_repo.sorted_commits_and_branches_per_repositories(
+        top_commit
+    )
+
+    # Print authentication repository status
+    print(f"Authentication Repository: {path}")
+    print(f"Head Commit: {head_commit}")
+    print("\nTarget Repositories Status:")
+
+    output: Dict = defaultdict(dict)
```
renatav, line 703: Since this code is already in `list_targets` , you can call it instead of copy-pasting, then just put everything together

taf/api/targets.py
```diff
@@ -662,3 +662,83 @@ def _update_target_repos(
...
+    # Load repositories and target repositories data
+    repositoriesdb.load_repositories(auth_repo)
+    target_repositories = repositoriesdb.get_deduplicated_repositories(auth_repo)
+    repositories_data = auth_repo.sorted_commits_and_branches_per_repositories(
+        top_commit
+    )
+
+    # Print authentication repository status
+    print(f"Authentication Repository: {path}")
+    print(f"Head Commit: {head_commit}")
```
renatav, line 700: For the authentication repository, we also want to print some of the same things like when processing target repositories: 
- bare
- up-to-date
- something-to-commit

taf/api/targets.py
```diff
@@ -662,3 +662,83 @@ def _update_target_repos(
...
+
+    # Load repositories and target repositories data
+    repositoriesdb.load_repositories(auth_repo)
+    target_repositories = repositoriesdb.get_deduplicated_repositories(auth_repo)
+    repositories_data = auth_repo.sorted_commits_and_branches_per_repositories(
+        top_commit
+    )
+
+    # Print authentication repository status
+    print(f"Authentication Repository: {path}")
```
renatav, line 699: Here you can't use this path, because it might be `.`, but you can print the repository's name: `auth_repo.name` or full path `auth_repo.path` (here we resolve the path, so we know we won't print the relative path).

taf/api/targets.py
```diff
@@ -662,3 +662,83 @@ def _update_target_repos(
         path = targets_dir / target_repo_name
         path.write_text(json.dumps(data, indent=4))
         taf_logger.info(f"Updated {path}")
+
+
+@check_if_clean
+def taf_status(path: str, library_dir: Optional[str] = None) -> None:
```
renatav, line 668: In our system, we have an authentication and its target repositories, but also the concept of dependencies. Dependencies are other authentication repositories which are linked with the current authentication repository. This is specified in a file called `dependencies.json`. Just like we have `load_repositories` that is used here, we also have `load_depndencies` and you can then interate over them `child_auth_repos= repositoriesdb.get_deduplicated_auth_repositories( auth_repo, commits).values()`

Then, you'd need to call this recursively. But we want to show clearly which repository is the parent, so we want to indent properly. Since list targets already outputs a json, then we can add information about the current repository and if it has dependencies, add their information under the `dependencies` key. It's very unlikely that we'll have deep nestings.

taf/tools/targets/__init__.py
```diff
@@ -237,3 +240,22 @@ def update_and_sign(path, library_dir, target_type, keystore, keys_description,
...
+
+    @targets.command(context_settings=dict(
+        ignore_unknown_options=True,
+        allow_extra_args=True,
+    ))
+    @catch_cli_exception(handle=TAFError)
+    @click.option("--path", default=".", help="Authentication repository's location. If not specified, set to the current directory")
+    @click.option("--library-dir", default=None, help="Path to the library's root directory. Determined based on the authentication repository's path if not provided.")
+    @click.pass_context
+    def status(ctx, path, library_dir):
```
renatav, line 252: Since this command is related to the repository as a whole, let's move it to the `repo` module/command.

taf/api/repository.py
```diff
@@ -156,3 +157,54 @@ def _check_if_can_create_repository(auth_repo: AuthenticationRepository) -> bool
...
+    print(f"{indent_str}Head Commit: {head_commit}")
+    print(f"{indent_str}Bare: {auth_repo.is_bare_repository()}")
+    print(f"{indent_str}Up to Date: {auth_repo.synced_with_remote()}")
+    print(f"{indent_str}Something to commit: {auth_repo.something_to_commit()}")
+    print(f"{indent_str}Target Repositories Status:")
+    # Call the list_targets function
+    list_targets(path=path, library_dir=library_dir)
+
+    repositoriesdb.load_dependencies(auth_repo, library_dir=library_dir)
+    if auth_repo.dependencies:
```
renatav, line 198: We probably made this confusion, but you actually don't need `auth.dependencies` here. As you saw, this directly stores the information that is listed in `dependencies.json`. `load_dependencies` instantiates there repositories and you don't have to worry about correctly setting the path etc. So you can due something like:
```
dependencies  = repositoriesdb.get_auth_repositories(auth_repp, head_commit)
for dep_repo in dependencies.values():
....
```

I'm currently getting an instantiation error, but this should fix it and make it simpler. Tested on my end

renatav (approved):
Tested and works now. Great job!

Issue #393, opened 2024-03-05
Add git hooks to prevent committing/pushing invalid commits


renatav:
Manual updates of an authentication repository can leave it in an invalid state. We want to create a git hook which we would copy to authentication repositories which would call TAF's validation and prevent a push if it fails

renatav:
Implementing this should not require that much knowledge of how the framework works, so not a bad first issue

renatav:
Resolved by merging #423

renatav:
Reopening this issue because we should add this hook after a successful update if not already present. That would ensure that every user has it, not just the person who created the repository

PR #423, opened 2024-05-21
feat: on repo creation a git hook that validates repo before push

ronantakizawa:

When running taf repo create, a git hook that validates the repo before pushing is made. Resolves issue #393 


taf/api/repository.py
```diff
@@ -116,6 +117,42 @@ def create_repository(
         write=True,
         no_commit_warning=True,
     )
+
+    # Create the .git/hooks directory and add the pre-push hook script if it does not exist
+    hooks_dir = Path(auth_repo.path) / ".git" / "hooks"
+    hooks_dir.mkdir(parents=True, exist_ok=True)
+
+    pre_push_script = hooks_dir / "pre-push"
+    if not pre_push_script.exists():
```
renatav, line 126: I haven't tested yet, but let's move this to a separate folder first. We don't currently have similar resources that we copy, so let's create a new folder `resources` in `taf/taf` and then copy from there.

taf/api/repository.py
```diff
@@ -116,6 +117,42 @@ def create_repository(
...
+if [ $VALIDATION_STATUS -ne 0 ]; then
+  echo "TAF validation failed. Push aborted."
+  exit 1
+fi
+
+# Allow the push if validation passes
+exit 0
+"""
+
+        with open(pre_push_script, "w") as file:
```
renatav, line 146: We generally use `pathlib.write_text(content)` or `.write_bytes(content)` to write to files

taf/api/repository.py
```diff
@@ -116,6 +117,42 @@ def create_repository(
...
+            file.write(pre_push_script_content)
+
+        # Make the pre-push script executable
+        pre_push_script.chmod(0o755)
```
renatav, line 150: This is os-specific. Will probably be necessary to check which operating system the user has. We can expect a lot of our end user to have Windows. I'd make a separate function which handles this

renatav (changes requested):

taf/utils.py
```diff
@@ -309,6 +310,30 @@ def to_tuf_datetime_format(start_date, interval):
     return datetime_object.isoformat() + "Z"
 
 
+def set_executable_permission(file_path: Path) -> None:
+    """
+    Set executable permission for the given file path.
+    Handles cross-platform differences.
+    """
+    try:
+        if platform.system() == "Windows":
```
renatav, line 319: I tested on Windows and this failed, but there were no issues with the hook. I don't think setting the permission on Windows is necessary.

taf/resources/pre-push
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Path to the TAF CLI executable
+TAF_CLI="taf"
+
+# Run the TAF validation command
+$TAF_CLI repo validate
```
renatav, line 7: It can take a while to validate the repository starting from the beginning. The point here is to check if the new commits are valid before pushing. After each update, we store the last validated commit. You'll see that in the same directory as the authentication repository, there is a folder named after it, like `_law` where there is a `last_validated_commit` file which is updated every time the updater fetches new commits. It should be sufficient to validate from that commit. `validate` command has a `--from-commit` option. We could maybe add a flag, `--from-latest` and if it is provided when calling the command, call `validate_repository` with this commit. `AuthenticationRepository` class has a `last_validated_commit` that can be used instead of reading from this file directly. That would be a huge improvement of this hook from the usability perspective

renatav (changes requested):
Looks good! Notice one issue, worked after addressing it. One other suggestion, but I don't feel strongly about it. Just need `--from-latest`

renatav:
I was using this today and I think that it's also important to provide some feedback that validation is in progress. A lot of our commands automatically commit and push. Because push is using `subprocess` (`subprocess.run(command, **options)`),  all of the output we get when we run `git push` through the command line is not visible. Because validation is started from the first commit if there is no `last_validated_commit`, it can take a while. It would make sense to check if there is a git hook [here](https://github.com/openlawlibrary/taf/blob/a1c7cb73ae58664a203ff75be939a678d10bfc44/taf/git.py#L1375) and log "Validating and pushing"

renatav (approved):

PR #432, opened 2024-05-28
fix: fix taf repo create issues

### LINKED ISSUE #424: Create repository issues
Issues that occur when running `taf repo create`:

- When keystore is not specified, we get three nearly identical warning messages in a row:
```
Keystore path is not passed, using default one: C:\Users\DavidGreisen\programming\archive\openlawlibrary\keystore
Keystore not specified. Using default location C:\Users\DavidGreisen\programming\archive\openlawlibrary\keystore
Keystore path is not passed, using default one: C:\Users\DavidGreisen\programming\archive\openlawlibrary\keystore
```
- If the keystore is not specified and and not found at the default location, abort creation of a new repository
- If keys-description is not specified, an error is raised `AttributeError: 'RoleSetupParams' object has no attribute 'get'`.  This probably started happening when we added `attrs`. We would expect a prompt asking people to enter the information instead. Fix the error and test the prompt.
- `path` parameter should be a required path argument. Do not default to the current folder. Users can pass `.` if the want to.

ronantakizawa:

Fixes repeating error messages in taf repo create and manual entry of keys-description. Fix #424


taf/api/roles.py
```diff
@@ -585,6 +587,10 @@ def _initialize_roles_and_keystore(
             taf_logger.info(
                 f"Keystore not specified. Using default location {default_keystore_path()}"
             )
+            if not Path(default_keystore_path()).is_dir():
```
renatav, line 590: So this does raise an error like it's stated in the issue, but now that I'm testing this, I'm remembering that the idea was to generate the keystore files if the keystore does not exist. I think that in `create_repository`, if the keystore does not exist, we should ask the user if they want to generate new keystore or not.
For additional context - our production repositories currently share online keys (this is not necessarily the best practice, it's just way more convenient). This means that when David is creating a new repository, if the keystore is not found, he forgot to specify the keystore path. On the other hand, I envisioned this feature to be flexible enough so that you can create new keys, write them to disk, print them to the command line for the users to do whatever they want with them later, set up yubikeys etc.

renatav (commented):
I think that we should handle this slightly differently compared to what's written in the issue. There's also a requirement in the issue to turn `path` into a required argument. Do not default to the current folder. This should be really easy to do, just update the click command. It should raise an error if `path` is not specified. I also a discovered another issue with `repo create` while testing this and will create a separate issue. I think it might be a bit more complex to get to the bottom of that, so does not have to be resolved now

taf/api/roles.py
ronantakizawa: The code should already ask to generate keystore files. 
If no keystore directory or files are there, the command first generates keys-description.json through user inputs on the CLI, generates a keystore directory and puts it there, then check if other keystore files are there and if not will generate them. 

Screenshot 2024-05-28 at 4 54 44 PM

renatav: That indeed is what should happen. However, if you do not specify keystore location, we get this error: `An error occurred while creating a new repository: Default Keystore Path D:\oll\library\oll-test-repos\test\keystore doesn't exist`

The default location is a directory named `keystore` inside the current working directory. Keystore path can be specified in that `keys-description` file, so we shouldn't turn keystore into a required argument. If it's not specified anywhere, we should print an error saying something like "Please provide keystore path". Looking at this, I would actually completely remove the concept of default keystore location. A while back, we thought that we could look for these keys inside the home directory, but the users found that to be very confusing

ronantakizawa: Just added a new commit. 

With the new version keystore is not a required argument. If the keys-description file is provided, it gets the keystore directory from there. If the keys-description file isn't provided, a user can generate one via the CLI. 

When generating via CLI, it will prompt the user to enter the keystore directory. If they don't provide one, they get the error "Please provide keystore path." If they provided it but it doesn't exist, they get the error "Keystore path {keystore} does not exist."

If the keystore path is provided and exists but doesn't have the keystore files, the user can generate them via CLI.

2 questions:
1. If the keystore path is provided but doesn't exist, the user gets an error and the CLI command quits. Would it be better to create the directory if it doesn't exist? 
2. When a user generates a keys-description file via CLI, it doesn't get saved anywhere as a JSON file. Is this a problem?

renatav: Yes, I think that we should create the directory if it does not already exist. Keen in mind that if keys-description is provided, it might not contain a keystore property.
I think that the idea was to print the generated keys-description to the command line. A user can copy-paste that to a file if they want to.

ronantakizawa: Made fixes so it creates a directory. If only keys-description is provided and it doesn't point to a keystone, it outputs an error saying "Please provide keystore path."

taf/api/roles.py
```diff
@@ -438,9 +438,10 @@ def _enter_roles_infos(keystore: Optional[str], roles_key_infos: Optional[str])
             "Enter keystore location if keys should be loaded from or generated to keystore files (leave empty otherwise): "
         )
         if len(keystore):
-            if not Path(keystore).is_dir():
-                taf_logger.info(f"{keystore} does not exist")
-                keystore = None
+            keystore_path = Path(keystore)
```
renatav, line 441: I had to think about this and I'd say that the new requirements were in conflict with some of our old requirements, so this might require a more work. One old requirement was to have the option to generate new keys and print them instead of saving them to files. This was supposed to enhance security, since you'd copy that value to a secure password manager. At the moment, we store root and target keys on Yubikeys, and the remaining ones (snapshot and timestamp) everyone stores inside keystore files. From the security standpoint, it is better to generate new keys every time, but at the moment, that is not the case. We want TAF to be able to be flexible enough so that it's up to the user what they want to do with their keys.
So let's do the following:
- If keys-description, or `roles_key_infos` how we call it here is not provided, ask the user to enter it. Do not ask the user about the keystore path at that time and do not print it at the end of that `_enter_roles_infos` function
- Now that we have `roles_key_infos`, check if all keys should be loaded from/stored to yubikeys
  - If that is not the case, check if keystore path is available (either read from keys-description.json or the keystore parameter)
  -  if keystore path is not specified, ask the user if they want to save/load keys from keystore files (make sure that the users enter y/n here, repeat until they provide an answer, instead of defaulting to one of the two answers)
  - If they their answer if negative,  do not raise an error, assume that the user wants the keys to entered using and printed to the command line. This use case should be tested in more detail, but is not a priority
 - if the keystore path is provided (regardless of how it was provided), check if the directory exists. 
   - Ideally, we could also iterate through all roles whose keys are not stored on yubikeys and check if the corresponding keystore files exist. Names of keystore files are based on names of the roles, so this can be automated. If some files are missing, print a message saying that keys of that and that role will be generated and ask them to confirm that that is what they want. This can be a future enhancement, so we can just create an issue now
   - if the directory does not exist, print that as a warning (use taf_logger) and ask the user if they want to enter a different path or generate new keys
   - Finally, once we have a keystore path that will be used going forward and if the user entered keys description through the command line, append that keystore path and then print the whole json

ronantakizawa: I made the changes, but I had some questions:
1. If the user chooses NO to save/load keys from keystore files or choose a directory for the keystore files, can I use the current directory (Set keystore to ".")? Without a specified directory a user currently can’t generate keystore files.
2. If a keystore path is declared in keys-description and as a command argument and both are different, which should be taken?

renatav: if the user does not want to save/load keys from the keystore files, we should just print the keys or ask the user to copy/paste the key into their terminal. I don't think that we've ever used that feature (or at least not recently), but it should all still be supported. Since this might require further changes, let's save the keystore files to `./keystore` for now. We should comment out that question though, it would be confusing
If keystore path is specified in both places, I think that the command argument should be used.

renatav: Now that I'm testing this. Let's leave the question, I'll take this over and look into what's happening with printing/loading from the command line.

taf/api/roles.py
```diff
@@ -574,33 +565,73 @@ def _initialize_roles_and_keystore(
...
+                        input(
+                            "Do you want to save/load keys from keystore files? [y/N]: "
+                        )
+                        .strip()
+                        .lower()
+                    )
+                    if use_keystore in ["y", "n"]:
+                        break
+                if use_keystore == "y":
+                    keystore = input("Please provide keystore path: ").strip()
```
renatav, line 596: Add `(default ./keystore)`. Let's also word this as "enter" instead of "provide"

renatav (commented):
There is a function `_setup_keystore_key` where  
```
Generate new keys? [y/N]: 
Write keys to keystore files? [y/N]: 
```
are printed again. There is a `skip_prompt` parameter. We should call this `_setup_keystore_key` with it set to True when creating a new repository if the user already answered the questions and confirmed that they would like to generate new keys and save to keystore

taf/api/roles.py
ronantakizawa: I made the changes and omitted the extra key generation text. 
I still get this in the output: 
...root.pub is not a file!
...root is not a file!
...targets.pub is not a file!

The command makes those files by the end of the command.

renatav (approved):

Issue #407, opened 2024-03-20
Profile updater


renatav:
Updater's performance is the biggest issue right now

ronantakizawa:
The slow performance is coming from functions clone_remote_and_run_tuf_updater and clone_or_fetch_users_auth_repo in updater_pipeline.py, both are either cloning or fetching. 
<img width="1409" alt="Screenshot 2024-05-21 at 6 24 32 PM" src="https://github.com/openlawlibrary/taf/assets/71115970/bade845d-5320-41ac-876f-76e1c5684ad0">
![output](https://github.com/openlawlibrary/taf/assets/71115970/e99f77bc-3522-427f-84e7-4cc63563573d)

PR #434, opened 2024-05-31
fix: Make Updater Faster

ronantakizawa:

Make the Updater faster by running the first 2 functions concurrently. 


renatav (commented):
I like this idea of parallelizing certain tasks. However, keep in mind that these groups of tasks still need to be executed in a specific order relative to other steps. So maybe inside `steps` list we could have arrays, and everything inside that array would be executed concurrently.

We also cannot close the issue #407 because we still need to figure out if there is anything else that we can do to speed up the update process when we are pulling new changes. It would be really helpful if you could profile the updater when there are no changes to be fetched. So run it, then run again and profile. In case of City of San Mateo, it takes the updater 9s (on my machine) to report that there are no changes. We want to reduce that time. We are not expecting you to make any further improvements, just to looked for parts that seem unnecessary or particularly slow.

dgreisen:
To elaborate on Renata's comment, my expectation is that we should be able to get an "empty" update to take not much more time than a git fetch. I.e. Well under a second. 

Possible reasons I can imagine: 
- when we do our local clone of the auth repo to create our quarantine repo, we aren't doing --bare so it's having to set up the entire workspace on disk
- we are doing unnecessary work after the got fetch/git status determines there's nothing new
- fetch is all always downloading 100s of speculative branches because those (correctly) never make it out of quarantine so they are never persisted on disk locally

renatav:
Repositories which are cloned to temp are bate repositories, but we fetch all branches, even thought that is not necessary. What'd I'd like us to do first is determine how much time is spent running individual steps. We don't mind you taking a stab at any of the discovered issues if you want to, but for now, the idea is to just figure out what we need to focus on the most. I was planning to do some work on the updater this month

ronantakizawa:
Here's the runtime of each function after testing the updater on one of the repos:

Updating repository

Updating repository sanipueblo/law


Checking if local repositories are clean...

Step check_if_local_repositories_clean took 0.00 seconds
Step set_existing_repositories took 0.00 seconds

Cloning repository and running TUF updater...


Validation repository: cloning repository


Validation repository: trying to clone from git@github.com:sanipueblo/law.git


Validation repository: successfully cloned from git@github.com:sanipueblo/law.git


Running TUF validation of the authentication repository...


Successfully validated authentication repository sanipueblo/law

Step clone_remote_and_run_tuf_updater took 7.71 seconds

Validating out of band commit and update type

Step validate_out_of_band_and_update_type took 0.00 seconds

Cloning or updating user's authentication repository...

Step clone_or_fetch_users_auth_repo took 1.94 seconds
Step load_target_repositories took 0.12 seconds

Finished cloning target repositories

Step clone_target_repositories_to_temp took 12.83 seconds

Validating initial state of target repositories...


Checking initial state of repositories

Step determine_start_commits took 0.01 seconds
Step get_targets_data_from_auth_repo took 0.00 seconds
Step check_if_local_target_repositories_clean took 0.23 seconds
Step get_target_repositories_commits took 13.30 seconds

Validating target repositories...


Validation of target repositories finished

Step validate_target_repositories took 0.00 seconds
Step validate_and_set_additional_commits_of_target_repositories took 0.00 seconds
Step update_users_target_repositories took 0.06 seconds

Merging commits into target repositories...

Step merge_commits took 0.26 seconds
Step remove_temp_repositories took 0.03 seconds
Step set_target_repositories_data took 0.00 seconds
Step print_additional_commits took 0.00 seconds

Called LifecycleStage.REPO handler. Event: unchanged


WARNING: Config file /Users/ronantakizawa/Documents/projects/nyu/407/config.json not found. Scripts might not be executed successfully


WARNING: No config data. Scripts might not be executed successfully


Called LifecycleStage.UPDATE handler. Event: unchanged

  completed in 36 seconds

ronantakizawa:
clone_remote_and_run_tuf_updater, clone_remote_and_run_tuf_updater, and get_target_repositories_commits have the longest runtimes. 

Adding parallel processing within each function significantly reduced runtime.

taf/updater/updater_pipeline.py
```diff
@@ -343,7 +344,7 @@ def clone_remote_and_run_tuf_updater(self):
                 self.operation = OperationType.UPDATE
             else:
                 self.operation = OperationType.CLONE
-
+
```
renatav, line 347: You can install a VSCode extension which removes the whitespaces if you're using VSCode

renatav (commented):
It makes a lot of sense to parallelize clone/fetch of target repositories. It looks like this is a few seconds faster now! CI build failed because of black. Please run black formatting and update the changelog.

dgreisen:
for an "unchanged" update I'd expect something like this:

- check whether local repos are in a valid state and are at "last_validated_commit"
  - if not error out
- local clone of auth repo into quarantine
- git fetch origin/main (or whatever)
- git status
- no changes. done.

Some big items that I don't think we need to be doing:
- Step clone_remote_and_run_tuf_updater took 8.01 seconds
  - Validation repository: trying to clone from [git@github.com](mailto:git@github.com):mohicanlaw/law.git
  - Validation repository: successfully cloned from [git@github.com](mailto:git@github.com):mohicanlaw/law.git
  - Running TUF validation of the authentication repository...
- Step clone_target_repositories_to_temp took 18.68 seconds
- Step get_target_repositories_commits took 8.29 seconds
which would leave 4 seconds

what is this doing?
- Step clone_or_fetch_users_auth_repo took 3.40 seconds
if we can get rid of that, then we'd be down to .6 seconds - right where we want to be.

dgreisen:
> I think that what we could do here is add a flag, which would allow you to specify if all steps should be run or if we can exit early.

Default should be to exit early. I don't see a need for a flag for end users, but if you need one for testing, that is fine.

> * if there are no new authentication commits, there could still be new pushes to target repositories

**I'm pretty ok with not knowing if there are unauthenticated target commits at the end of the validated target commits. Definitely not worth pulling target commits just for that purpose.**

> * if these target repositories are not allowed to contain unauthenticated repositories, that could be indicative of malicious pushes, or an error on our end (like incomplete merge) which should be detected

**I think it's actually better that we don't know about these. that way a malicious person without access to key would be unable to break the update.**

> * if these repositories are allowed to contained unauthenticated commits, this is not an error. The updater used to merge this commits, but I believe that that is no longer the case, so it's only the previous scenario that is problematic

**correct it no longer does, now it just tells you there are commits we can merge. those are really just for the convenience of the developer, an end user should never be using them.**

* what is `clone_or_fetch_users_auth_repo` and how is that different from

renatav (approved):

renatav:
If there are invalid commits following valid ones, the updater will now partially update the repositiies. Of we were to ignore invalid additional commits compared to how uodater is currently handling them, only the update status would be different and the uodater would execute different scripts. When running the build command, a user would have to pull down new changes first. If we were to run validation from the first commit preceeding the first newly built auth commit, the validation would fail. If we ran it from the first new commit, it would be successful. I'm concerned about the build workflow, since an advanced user, like Tom, could just manually pull if running the uodater does not allow them to run build. This can all easily happen if our CI for some reason partially merged commits. Maybe the answer here is to move some of those responsibilities to platform if we do not want the uodater to provide such information to the end-users. We've had too many issues with the repositories caused by invalid commits, especially if the are not the newest ones

PR #435, opened 2024-05-31
Fix generation of keys when they should be printed to the command line

### LINKED ISSUE #433: Create repo error when generating new keys
We made a lot of updates to the signing process, and it's possible that this error started happening as a result of some of those reworks. This might only happen if the users does not want to store the keys in a keystore file. An error occurs while setting up roles.

renatav:

A feature that we rarely use, but still want to support is printing keys to the command line instead of storing them to the keystore files. This enables the users to copy them to a secure key manager. 

Fix #433

### DIFF
--- a/taf/api/roles.py
+++ b/taf/api/roles.py
@@ -542,7 +542,7 @@ def _initialize_roles_and_keystore(
     roles_key_infos: Optional[str],
     keystore: Optional[str],
     enter_info: Optional[bool] = True,
-) -> Tuple[Dict, str, bool]:
+) -> Tuple[Dict, Optional[str], bool]:
     """
     Read information about roles and keys from a json file or ask the user to enter
     this information if not specified through a json file and enter_info is True.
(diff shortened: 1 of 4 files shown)


taf/api/roles.py
```diff
@@ -580,64 +580,58 @@ def _initialize_roles_and_keystore(
...
+                )
+                if use_keystore in ["y", "n"]:
+                    break
+            if use_keystore == "y":
+                keystore = (
+                    input("Enter keystore path (default ./keystore): ").strip()
+                    or "./keystore"
+                )
+            else:
+                taf_logger.info("Keys will be entered/printed from/to command line")
```
n-dusan, line 600:
```suggestion
                taf_logger.info("Keys will be entered and then printed from the command line...")
```

taf/api/roles.py
```diff
@@ -580,64 +580,58 @@ def _initialize_roles_and_keystore(
...
-                print(
-                    "Assuming the keys will be entered using and printed to the command line."
-                )
-                keystore = "."
+                print(f"Created keystore directory at {keystore}")
                 skip_prompt = True
                 break
+            else:
+                enter_new_path = (
+                    input("Do you want to enter a different path? [y/N]: ")
```
n-dusan, line 625:
```suggestion
                    input("Do you want to enter a different path to the keystore? [y/N]: ")
```

taf/api/roles.py
```diff
@@ -580,64 +580,58 @@ def _initialize_roles_and_keystore(
...
+                    .strip()
+                    .lower()
+                )
+                if enter_new_path == "y":
+                    keystore = input("Enter a different keystore path: ").strip()
```
n-dusan, line 630:
```suggestion
                    keystore = input("New keystore path: ").strip()
```

n-dusan (approved):
Approved with some minor wording comments (no need to request re-review)!

Issue #379, opened 2024-01-16
Add support for Python 3.11, 3.12


n-dusan:
This PR added support for Python 3.10 . I doubt that much more will need to be done: https://github.com/openlawlibrary/taf/pull/247
It is essential that we retain compatibility with older versions of Python

Rana-KV:
Hi @renatav, I would like to solve this issue. Can you assign it to me?

PR #440, opened 2024-06-06
Update to support Python 3.11 and 3.12

Rana-KV:
## Description

**Addresses Issue**: #379
fixes: #379
**Summary of Changes**:
1. **Dependency Updates**:
   - Updated dependency versions to support Python 3.11 and Python 3.12.
   - The build process will now consider the Python version to select the appropriate libraries.
   - **Click Library**: Updated to version 8.*.*.
   - **pygit2 Library**:
     - For Python 3.8, updated to version 1.10.1.
     - For Python 3.9 and 3.10, updated to version 1.13.3.
     - For Python 3.11 and 3.12, updated to version 1.14.1.

2. **Build and Workflow Updates**:
   - Updated GitHub workflows to accommodate the new Python versions and dependencies.
   - Updated Azure Pipelines configuration for compatibility with the updated dependencies and Python versions.

3. **Documentation**:
   - Updated `CHANGELOG.md` to reflect these changes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 from setuptools import find_packages, setup
 from importlib.util import find_spec
+import sys
 
 PACKAGE_NAME = "taf"
 VERSION = "0.29.3"
@@ -46,6 +47,14 @@ tests_require = [
 
 yubikey_require = ["yubikey-manager==4.0.*"]
 
+# Determine the appropriate version of pygit2 based on the Python version
+if sys.version_info >= (3, 9):
+    pygit2_version = "pygit2==1.14.1"
+elif sys.version_info >= (3, 8) and sys.version_info < (3, 9):
+    pygit2_version = "pygit2==1.13.3"
+elif sys.version_info >= (3, 7) and sys.version_info <= (3, 10):
+    pygit2_version = "pygit2==1.10.1"
+
 kwargs = {
     "name": PACKAGE_NAME,
     "version": VERSION,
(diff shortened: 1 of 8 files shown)


renatav (commented):
Looks like flake 8 checks are more thorough for Python 3.12, so the build failed. See https://github.com/openlawlibrary/taf/actions/runs/9407771803/job/25914281561?pr=440
Could you take a look at those issues. They seem trivial.

We also always update this [CHANGELOG ](https://github.com/openlawlibrary/taf/blob/master/CHANGELOG.md) file before merging a PR which introducing changes worth noting.

Rana-KV:
Hi @renatav, I have fixed few of the errors and added trivial errors (like space after ':' ) in ignore list.
I have updated the [CHANGELOG.md](https://github.com/openlawlibrary/taf/blob/master/CHANGELOG.md) in the commits mentioning that Python 3.11 and 3.12 are supported. Should I keep them or remove it?

setup.py
```diff
@@ -46,6 +47,14 @@ def finalize_options(self):
 
 yubikey_require = ["yubikey-manager==4.0.*"]
 
+# Determine the appropriate version of pygit2 based on the Python version
+if sys.version_info >= (3, 9):
```
renatav, line 51: Had to check something, but there was a reason why `pygit2` was pinned to `1.9`. A git security check was added in 1.9.1,  which broke something in our system. It'll need to stay pinned. Since our system is still on Python 3.10, we can set different versions of `pygit2` in case of 3.11 and 3.12 if that is necessary

renatav (changes requested):

renatav:
@Rana-KV We definitely want to add this to the changelog. Could you please also link a PR there? Take a look at other changelog entries. We sort PRs in the descending order. Thanks for looking into the flake issues!

setup.py
Rana-KV: Sure, I can pin the version for `Python 3.10`. But the testcases used in pytest didn't raise any error when I ran the TAF builds with latest versions of `pygit2` for `Python 3.10`. On the other side `Python 3.11` needs `pygit2 version >=1.11.x` and `Python 3.12` needs `pygit2 version >=1.13.x` 

Should I just change it for  `Python 3.10` ?

renatav: Let's pin it for `Python <= 3.10`. The error is not happening in TAF, it's in a component of our system which uses TAF. We'll update it if once we've addressed the core issue.

renatav (approved):

Issue #430, opened 2024-05-23
Require target-name parameter 


renatav:
`taf targets add-repo --target-name` parameter should be a required path argument

Rana-KV:
Hi @renatav, I would like take up this issue and work on it.

renatav:
@Rana-KV Awesome, I assigned it to you,

Rana-KV:
Thanks for assigning the issues @renatav
I have question regarding the changes. If target-name is required, then can I remove the fallback option to use target-path?

![TAF-Capture](https://github.com/openlawlibrary/taf/assets/19876949/f04ab22f-fcd8-404e-a816-caf87593b403)

renatav:
@Rana-KV We tend to follow a certain convention regarding where the repositories are located. Their filesystem paths tend to include their names. This will be the case if you use the updater to clone the repositories, like it is intended. So, if you have a name, the chances are that we can also correctly calculate the path. If for whatever reason that is not the case, we probably should still make it possible to specify a different path. So, let's actually reverse the order of this `if` and `elif` instead of removing the `elif`

Rana-KV:
Thanks for the information @renatav, I have reversed the order and still kept the description in targest.py about target-name parameter as optional because it be still done with just taget-path.

PR #442, opened 2024-06-08
Fix: Update to repository addition order

Rana-KV:
## Description
fixes: #430

Updated the order of adding repository based on target_path if provided, else based on target_name if target_path is empty. Completely making target_name mandatory might break the functionality of updater based repository additions, so did not completely remove the target_path based repository addition.

Updated the `--target-name` option as required in the CLI. However, I have maintained the default assignment as `None` because the API function relies on `None` to determine if the `target-name` was provided or not.

### DIFF
--- a/taf/tools/targets/__init__.py
+++ b/taf/tools/targets/__init__.py
@@ -42,7 +42,7 @@ def add_repo_command():
         `namespace1\\repo1`, the target's path will be set to `E:\\examples\\root\\namespace1\\repo1`.""")
     @catch_cli_exception(handle=TAFError)
     @click.option("--path", default=".", help="Authentication repository's location. If not specified, set to the current directory")
-    @click.option("--target-name", required=True, default=None, help="Namespace prefixed name of the target repository")
+    @click.argument("target-name")
     @click.option("--target-path", default=None, help="Target repository's filesystem path")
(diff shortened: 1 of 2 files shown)


renatav (commented):
This looks good, but it's also necessary to update the CLI command. `target-name` should be a required argument instead of an option

Rana-KV:
I have added the `--target-name` option as required in the CLI. However, I have maintained the default assignment as `None` because the API function relies on `None `to determine if the `target-name` was provided or not.

taf/tools/targets/__init__.py
```diff
@@ -42,7 +42,7 @@ def add_repo_command():
         `namespace1\\repo1`, the target's path will be set to `E:\\examples\\root\\namespace1\\repo1`.""")
     @catch_cli_exception(handle=TAFError)
     @click.option("--path", default=".", help="Authentication repository's location. If not specified, set to the current directory")
-    @click.option("--target-name", default=None, help="Namespace prefixed name of the target repository")
+    @click.option("--target-name", required=True, default=None, help="Namespace prefixed name of the target repository")
```
renatav, line 45: We typically use an argument instead of an option that needs to be provided. See other commands. I'd like to maintain consistency, so I'd rather use it here as well.

Rana-KV: Understood, sorry for the confusion. I was looking use the same declaration. I have now updated it properly.

renatav (approved):
Looks good now!

Issue #410, opened 2024-03-28
Transition to `yubikey-manager` 5.x


n-dusan:
I couldn't sign a repository using YubiKey 5C with `yubikey-manager` 4.0.9. 
Look into upgrading our `cryptography` dependency as well

Rana-KV:
Hi @n-dusan,

I've noticed that updating to YubiKey Manager 5.x introduces significant changes to the project due to deprecated functions and module restructuring. For example, `connect_to_device` from ykman.device no longer exists. Adapting the code to support YubiKey Manager 5.x would make it incompatible with previous versions of the project.

One approach to address this could be to check the version numbers and update the code to work with both versions, although this might be less efficient. This would ensure backward compatibility while taking advantage of the new functionalities provided by YubiKey Manager 5.x.

Let me know your thoughts on this or if you have any other suggestions.

renatav:
@Rana-KV Thanks for looking into this issue, this is currently a priority for us. I don't think that backward compatibility is something we need to worry about here. We have API functions for signing and setting up YubiKeys, so no one should be using any of the impacted code directly. The reason why we need the do this is because we recelty received a donation of new 20 YubiKeys from Yubico and we can't use them. We only need to make sure that you'll still be able to sign with an older YubiKey after the update, so we'll need to test that before merging

PR #444, opened 2024-06-09
Support Yubikey Manager 5.1.x

Rana-KV:
## Description
fixes: #410 
- Updated code to support Yubikey Manager 5.1.x 
- Changes are made to remove deprecated modules and functions and use newer ones.
- Dropped support for Yubikey Manager 4.x

### DIFF
--- a/taf/yubikey.py
+++ b/taf/yubikey.py
@@ -14,7 +14,7 @@ from cryptography.hazmat.primitives.serialization import load_pem_private_key
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import rsa, padding
 from tuf.repository_tool import import_rsakey_from_pem
-from ykman.device import list_all_devices, connect_to_device
+from ykman.device import list_all_devices
 from yubikit.core.smartcard import SmartCardConnection
 from ykman.piv import (
     KEY_TYPE,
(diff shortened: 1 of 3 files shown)


Rana-KV:
I have to update the Changelog document, I want to update it after the review and thought the other PR might have merge conflicts with this one.

taf/yubikey.py
```diff
@@ -110,31 +110,34 @@ def _yk_piv_ctrl(serial=None, pub_key_pem=None):
...
+                    or device_pub_key_pem[:-1] == pub_key_pem
+                ):
+                    break
+                yield session, info.serial
     else:
-        connection, _, device = connect_to_device(serial, [SmartCardConnection])
-        session = PivSession(connection)
-        yield session, device.serial
+        for dev, info in list_all_devices():
+            if info.serial == serial:
```
renatav, line 135: Tested and debugged, here we need `if serial is None or info.serial == serial:` I was getting an error because nothing was getting yielded. At the moment, we call this without specifying the serial. I believe that this was was supposed to let us use multiple yubikeys at the same time, but we never managed to finish that work. It actually should be much easier now wit the new yubikey manager, but definitely not related to this PR. With that change, I was able to set up a new Yubikey and sign!

renatav (commented):
Thank you for looking into this one, this was a priority for us! I found one small issue, worked following that change.
Any idea why it's saying that `oll-bot` committed this?

Rana-KV:
> Thank you for looking into this one, this was a priority for us! I found one small issue, worked following that change. Any idea why it's saying that `oll-bot` committed this?

I'm sorry, I believe I have mistakenly pushed changes and commit from my test setup, which runs GitHub Actions. One of these actions sets the git config username to `oll-bot`. I forgot about this configuration and pushed the code.

taf/yubikey.py
Rana-KV: Yup, I think this code won't allow to use Yubikey with unknown public key or device serial number. It makes sense to add the line of code you suggested. I have made the changes and committed to the PR. Thanks for catching it.

We can expand to support multiple key support, definitely something interesting to work in the next weeks.

renatav:
@Rana-KV You can update the commit history and force push to your branch, you don't have to create a new PR. Should not be too difficult. And then just update the changelog and we can get this merged in!

Rana-KV:
@renatav, Thanks I have did the needed changes.

renatav (commented):
Looks good! Could you please update the changelog?

renatav (approved):

Issue #367, opened 2023-09-15
Package taf into standalone executable for Windows/Linux/macOS


renatav:
Stelae needs to include TAF, so this also means that we need to figure out how to package TAF

renatav:
Issue assigned to @ronantakizawa (cannot assign issues directly)

PR #447, opened 2024-06-12
feat: new github action to create standalone executable of TAF

ronantakizawa:

Added a new Github action to build standalone executables of TAF. Resolves issue #367 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,13 +1,19 @@
-on: [push, pull_request]
+on:
+  push: {}
+  pull_request:
+    types: [closed]
+  release:
+    types: [published]
 
 jobs:
-  build:
+  run-tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v3
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v3
         with:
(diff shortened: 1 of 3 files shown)


.github/dockerimages/Dockerfile.macos
```diff
@@ -0,0 +1,76 @@
+# Use the latest macOS compatible image (experimental)
+FROM bot62726170686f67/macos-cross-rust:latest
```
renatav, line 2: This image is quite old, can we use a newer one? Are we using Rust for anything here? Or will we need it in the future?

.github/dockerimages/Dockerfile.macos
```diff
@@ -0,0 +1,76 @@
...
+    make install
+
+# Set environment variables for OpenSSL
+ENV LDFLAGS="-L/home/arch/openssl/lib"
+ENV CPPFLAGS="-I/home/arch/openssl/include"
+ENV PKG_CONFIG_PATH="/home/arch/openssl/lib/pkgconfig"
+ENV PATH="/home/arch/openssl/bin:$PATH"
+ENV LD_LIBRARY_PATH="/home/arch/openssl/lib:$LD_LIBRARY_PATH"
+
+# Download and install Python 3.8 from source with OpenSSL and enable shared library
```
renatav, line 20: Same question, can we use a newer Python?

.github/workflows/ci.yml
```diff
@@ -1,29 +1,120 @@
...
           git config --global user.name oll-bot
           git config --global user.email developers@openlawlib.org
+
       - name: Run pre-commit and test with pytest
         run: |
           pre-commit run --all-files
           pytest taf/tests
+
+      - name: Build standalone executable with Docker (Linux)
+        if: matrix.os == 'ubuntu'
```
renatav, line 45: Question: does the host OS matter if we are using docker containers?

.github/workflows/ci.yml
```diff
@@ -1,29 +1,120 @@
...
+        run: |
+          docker build -t taf-macos -f .github/dockerimages/Dockerfile.macos .
+          docker cp $(docker create --rm taf-macos):/usr/src/app/taf/dist/taf taf-macos
+
+      - name: Test standalone executable (Linux)
+        if: matrix.os == 'ubuntu'
+        id: test-linux
+        run: |
+          ./taf-linux --help > actual_output.txt
+          if ! diff -q actual_output.txt expected_output.txt; then
```
renatav, line 67: My concern here is that we would have to update expected output every time we change the cli (add, remove, rename a command group).  Just this part should be sufficient if we are checking if it was installed:
```
Usage: taf [OPTIONS] COMMAND [ARGS]...

  TAF Command Line Interface
```
We could discuss is there is more that we should test, but that does not have to be a part of this PR

renatav (commented):
There seems to be an issue with actions: https://github.com/openlawlibrary/taf/actions/runs/9490478776

I'm thinking for now, we could build when a PR gets merged, so that we can properly tests this. But I'd like us to then build only if we release. We do not want to build the executable on push

.github/workflows/ci.yml
```diff
@@ -1,29 +1,114 @@
...
+  pull_request:
+    types: [closed]
+    branches:
+      - main
+  release:
+    types: [published]
 
 jobs:
-  build:
+  build-and-test:
```
renatav, line 10: We want to run the old code every time we push, and everything related to executables only in these situations. I think that we can accomplish this by separating the jobs. But, before merging, let's also make sure that this can be executed successfully, so run the build on push too, then revert that before merging.

.github/workflows/ci.yml
```diff
@@ -1,29 +1,123 @@
...
         run: |
           git config --global user.name oll-bot
           git config --global user.email developers@openlawlib.org
+
       - name: Run pre-commit and test with pytest
         run: |
           pre-commit run --all-files
           pytest taf/tests
+
+  build-and-test-executables:
```
renatav, line 44: We could run this only if the the previous job is successful. That way, we won't build the executables if tests fail for some reason. You could do that like this:
```
    needs: run-tests
    runs-on: ubuntu-latest
    if: ${{ needs.run-tests.result == 'success' }}
```
If the first job is named `run-tests`

renatav (changes requested):
I created a branch based on your branch and made the changes I suggested here, which then triggered the build. This is what happened: https://github.com/openlawlibrary/taf/actions/runs/9522838340

Feel free to push freely until the build passes, then remove the test commits. Before merging, we can add the condition to only build on merge to master and release

renatav (approved):

renatav (approved):

PR #453, opened 2024-06-21
Rework updater testing framework

### LINKED ISSUE #383: Rework updater testing framework
At the moment, for each test we have a set of git repositories which were set up manually and committed after renaming the `.git` folders to `git`. Given the tools that were developed after initial this testing framework was set up, it should be possible to recreate all of the repositories and the appropriate states they are programmatically. This should make it possible to add new tests with much less effort.

List of tests that we need to add:
- Execution of scripts
- Update of repositories which reference other repositories (which contain `dependenceis.json`)
- Add tests for more states in which authentication and target repositories can be in when the updater is called. See issue #210 
which lists a number of different states the repositories can be in.
- Add tests for invalid metadata - `root`, `targets`, `snapshot`, `timestamp`

We also need to add tests which would include authentication repositories which contain dependencies and scripts.

### LINKED ISSUE #406: Tests are failing on Windows after fresh install
There used to be just one `conftest` file before we create one per test package. These inner `conftest` files reference the top-level `conftest`. That works on CI and it used to work on my old virtual env, but it started erroring out. Works if that top-level `conftest` is removed. We could still import constants and everything other than fixtures. Fixtures can be moved to the inner `conftest` files. Is there a better way to fix this without this rework?

Might be as simple as bumping pytest version

renatav:

Complete rework of the updater testing framework. Instead of using previously set up test repositories (we used to create those manually and then rename `.git` directory to `git` to avoid having submodules), repositories are set up programmatically at the beginning of each test. Since we now have the functions for creating and updating repositories, this is cab be accomplished quite easily.

At the beginning of each test, an authentication repository is set up. The way in which the authentication repository can be set up can be set by using `setup_type`. For example, it is possible to create a repository with and without `info.json`. It is also configured which target repositories should be linked with that auth repo and if they can contain unauthenticated commits. E.g.

```@pytest.mark.parametrize(
    "origin_auth_repo",
    [
        {
            "setup_type": "all_files_initially",
            "targets_config": [{"name": "target1"}, {"name": "target2"}],
        },
        },
        {
            "setup_type": "no_info_json",
            "targets_config": [{"name": "target1"}, {"name": "target2"}],
        },
```

In every test, we can use the `origin_auth_repo` fixture which is set up in accordance with these specifications. It and the reference target repositories can then be updated easily by using the `SetupManager`. E.g.

```
    setup_manager = SetupManager(origin_auth_repo)
    setup_manager.add_task(add_unauthenticated_commits_to_all_target_repos)
    setup_manager.add_task(add_valid_target_commits)
    setup_manager.execute_tasks()
```
By using this `SetupManger`, we parameters of setup task like auth repo and target repositoires are automatically set. In this example, we ad unauthenticated commits, followed by signed commits, aiming to test if the updater will correctly fail and report an mismatch between commits. So, before running update, we use this to create the state we want to test. `add_unauthenticated_commits_to_all_target_repos` and `add_valid_target_commits` are two of several functions that were added and can be used to update the repositories. They rely on TAF's api.
After the repositories have been set up, we can also perform other actions in order to reproduce certain states, like run the updater and clone the repositories, before running it again, or clone just one of the repositories without the updater prior to invoking the updater. There operations are performed without the `SetupManager`, it only updates origin repos.

There already was a test for updating a repository with dependencies which was set up programmatically. It was updated, but not completely reworked. It can be made more flexible in the future, that could be a separate PR.

Removed git repositories we used to use for testing. We can go back through history and see which one existed if we needed to reference that.

Update `pytest` version and dropped `python 3.7` support since the new pytest does not support it.

Fixes #383 
Fixed #406 


taf/tests/test_updater/conftest.py
```diff
@@ -143,6 +189,40 @@ def sign_target_files(library_dir, repo_name, keystore):
...
+    elif setup_type == "repositories_and_mirrors_added_later":
+       return setup_repository_repositories_and_mirrors_added_later(
+            repo_name, targets_config, is_test_repo
+        )
+    elif setup_type == "no_target_repositories":
+       return setup_repository_no_target_repositories(
+            repo_name, targets_config, is_test_repo
+        )
+    else:
+        raise ValueError(f"Unsupported setup type: {setup_type}")
```
n-dusan: nitpick: maybe this should be `setup_state` instead of `setup_type`?

suggestion: I think we need documentation for these states. One way is to document each individual function we call out (e.g. `setup_repository_no_target_repositories`). The other way would be to document the `setup_type` DSL values? I'm leaning towards `setup_types`. That way, someone looking at the `setup_types` would know what each of them is meant to do. What do you think about:

```python
class SetupState(enum.Enum):
    # All target files are initially present in the authentication repository.
    # The files meant to be included are:
    #   - protected/info.json
    #   - mirrors.json
    #   - repositories.json
    # ...  
    ALL_FILES_INITIALLY= "all_files_initially"
    # Initialized without the `protected/info.json` file
    # ... (more documentation if necessary)
    NO_INFO_JSON= "no_info_json"
    ...
```
That way, aside from the centralized state, other contributors can reason about the way current states are intended to work. Is there a better way of doing it? What do you think?

taf/tests/test_updater/test_update/test_update.py
```diff
@@ -8,174 +8,257 @@
...
+        client_dir,
+    )
+
+
+
+@pytest.mark.parametrize(
+    "origin_auth_repo",
+    [
+        {
+            "date": "2020-01-01",
```
n-dusan: question: What does the `date` attribute do? Maybe a comment in the `SetupManager` would be helpful

taf/tests/test_updater/test_update/test_update_invalid.py
```diff
@@ -0,0 +1,52 @@
...
+from taf.tests.test_updater.update_utils import check_if_last_validated_commit_exists, clone_repositories, update_and_check_commit_shas, update_invalid_repos_and_check_if_repos_exist
+from taf.updater.types.update import OperationType, UpdateType
+
+
+def setup_module(module):
+    settings.update_from_filesystem = True
+
+
+def teardown_module(module):
+    settings.update_from_filesystem = False
```
n-dusan: suggestion: Could we centralize `setup_module` and `teardown_module`? I see we're using them in all tests

taf/tests/test_updater/test_clone/test_clone_invalid.py
```diff
@@ -111,11 +112,13 @@ def test_clone_invalid_target_invalid_metadata(origin_auth_repo, client_dir):
     setup_manager.add_task(swap_last_two_commits)
     setup_manager.execute_tasks()
 
+    # On CI, the error message is not always the same, but always reports that the metadata is invalid
+    # so use a more generic error pattern for now
     update_invalid_repos_and_check_if_repos_exist(
         OperationType.CLONE,
         origin_auth_repo,
         client_dir,
-        INVALID_VERSION_NUMBER_PATTERN,
```
n-dusan: question: `INVALID_VERSION_NUMBER_PATTERN` is now an unused global variable. Is that intended? If so, let's check if any other variables are unused and remove them

n-dusan (changes requested):
Incredible that ~67,000 lines are removed, must've felt really nice!

I ran the new tests locally and they work. I've left mostly stylistic comments, so once we resolve, add a Changelog entry and we'll be ready to merge

taf/log.py
```diff
@@ -20,13 +20,19 @@
 
 
 def disable_console_logging():
-    taf_logger.remove(console_loggers["log"])
-    disable_tuf_console_logging()
+    try:
+        taf_logger.remove(console_loggers["log"])
+        disable_tuf_console_logging()
+    except ValueError:
+        pass
```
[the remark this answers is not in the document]
renatav: If you call it twice, meaning that it was already removed

taf/tests/test_updater/test_update/test_update.py
renatav: It allows you to modify the current time using `freezetime` and create repositories with expired metadata

taf/tests/test_updater/conftest.py
renatav: If enum can be used in `parameterize`, then it makes more sense to use an enum. I don't feel strongly about documenting these, the names are descriptive enough and some understanding of the framework is necessary, but I'll add more notes if you think they would be helpful

n-dusan (approved):

Issue #397, opened 2024-03-14
Add opt-out options to `taf repo validate`


n-dusan:
Options to add to taf validation: don't validate targets, don't validate targets, don't download upstream.

- #450
- --no-targets
- --no-upstream

Issue #450, opened 2024-06-14
Add `--no-deps` to taf repo `update`, `clone` and `validate`


n-dusan:
The `--no-deps` flag would then only update the current repository, and not traverse/update the dependent repositories from `dependencies.json`

renatav:
Closed by merging #455

PR #455, opened 2024-06-24
feat: three new flags added to TAF repo (397, 450)

jc12690:
fixes: #397, #450 
- Added --no-targets flag allowing users to skip target repository validation when validating the authentication repo
- Added --no-upstream flag allowing users to skip upstream comparisons
- Added --no-deps flag allowing users to only update the current repository and not update dependent repositories from dependencies.json


taf/updater/updater.py
```diff
@@ -186,6 +186,21 @@ class RepositoryConfig:
         default=False,
         metadata={"docs": "Whether update fails if a warning is raised. Optional."},
     )
+    # JMC: Addition of --no-deps option to disallow updating of the last validated commit if any part of validation is skipped
```
renatav, line 189: Probably don't need any additional comment here because the metadata is descriptive enough. In this case, it's more accurate than this comment

taf/updater/updater.py
```diff
@@ -186,6 +186,21 @@ class RepositoryConfig:
...
+        metadata={"docs": "Specifies whether or not to update dependencies. Optional."}
+    )
+    # JMC: Addition of --no-targets option to allow user to skip target repos when validating the authentication repository.
+    no_targets: bool = field(
+        default=False,
+        metadata={"docs": "Flag to skip target repositiory validation and validate only authentication repos. Optional."}
+    )
+    # JMC: Addition of --no-upstream option to allow user to opt out of comparing with the remote repository and be added to clone and update
+    no_upstream: bool = field(
+        default=False,
```
renatav, line 201: This feature was not very well described in the issue. What this is supposed to do is only opt out of comparing target repositories with upstream, if case there were no changes pushed to the authentication repository. If there were no changes pushed to the auth repo, then we are sure that there are no valid changes pushed to the target repositories and we are sure that nothing will be pulled down. However, by comparing the local target repositories with the remote repositories, we can detect errors. Only some users need to do this. This should be set too `True` by default. When we only want to validate local repositories, we can run `taf repo validate`

taf/updater/updater.py
```diff
@@ -366,6 +385,9 @@ def _update_named_repository(
     scripts_root_dir=None,
     checkout=True,
     excluded_target_globs=None,
+    no_deps=False,
+    no_targets=False,
+    no_upstream=False,
```
renatav, line 390: Set to `True`

taf/updater/updater.py
```diff
@@ -425,6 +447,18 @@ def _update_named_repository(
     if url in visited:
         return
     visited.append(url)
+
+    # JMC: Added logic for no_upstream
+    if no_upstream:
```
renatav, line 452: This is probably not what we want to do here, please take a look at the previous comment. The issue lacked some important information.

taf/updater/updater.py
```diff
@@ -463,101 +497,105 @@ def _update_named_repository(
...
+                _merge_commit(auth_repo, auth_repo.default_branch, last_commit, True)
+                # update the last validated commit
+                if not excluded_target_globs:
+                    auth_repo.set_last_validated_commit(last_commit)
+
+            # do not call the handlers if only validating the repositories
+            # if a handler fails and we are in the development mode, revert the update
+            # so that it's easy to try again after fixing the handler
+            # JMC: Added "and not no_targets:" to this first line of code
+            if not only_validate and not excluded_target_globs and not no_targets:
```
renatav, line 581: This is correct, but in order to actually avoid update of target repositories and support `no-upstream` flag as well, you'll need to update `updater_pipeline` That's why that issue did not have the "good first issue label".  Handling `no-targets` will probably be pretty easy.  We can skip steps which download and validate target repositories. You'll see that there is an enum called `RunMode` which allows us to configure which steps we'll run depending on if we're running update or local validation. At the moment, we have this:

```
steps=[
                (self.set_existing_repositories, RunMode.UPDATE),
                (self.check_if_local_repositories_clean, RunMode.UPDATE),
....

So, we should somehow extent this so that we can specify that a step should be skipped if `no_targets` is set to `True`. We also have `no_upstream`  which is a bit trickier because we can't know if we should skip update of target repositories until we've executed `clone_or_fetch_users_auth_repo`

An idea that came to my mind now is that we could add another (optional) entry to these tuples, and that would be a method that would be called to check if the step should be executed.

```
(load_target_repositories, , RunMode.ALL, self.check_if_targets_skipped)

def check_if_targets_skipped(self):
  return self.no_targets or (self.no_upstream and self.auth_repo_not_updated)
```
`auth_repo_not_updated` would be set after auth repo validation.
So with those update, updating  this for loop `for step, step_run_mode in self.steps:` should be sufficient

renatav (approved):

PR #458, opened 2024-06-28
Move release CI from Azure Pipelines to GitHub Actions

### LINKED ISSUE #367: Package taf into standalone executable for Windows/Linux/macOS

renatav:

- Implement deployment of wheels to PyPi using GitHub actions, instead of Azure
- Upload executables as release assets
- Confirmed that TAF works on Windows and MacOS without copying DLLs. This means that platform-specific wheels are not longer necessary

- Successful release [build](https://github.com/openlawlibrary/taf/actions/runs/9707362671)
- Executables deployed, see this [release](https://github.com/openlawlibrary/taf/releases/tag/test4)
- Wheels uploaded to Test PyPi. Install like this: `pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ test-taf[yubikey]`

Fixes #367


.github/workflows/ci.yml
```diff
@@ -41,21 +44,84 @@ jobs:
           pre-commit run --all-files
           pytest taf/tests
 
+  build_and_upload_wheel:
+    runs-on: ubuntu-latest
+    needs: run_tests
+    if: github.event_name == 'release'
```
n-dusan, line 50: Should this be on release or on tag?

.github/workflows/ci.yml
```diff
@@ -41,21 +44,84 @@ jobs:
...
+          pip install packaging
+          pip install twine==3.8.0
+
+      # - name: Upload to TestPyPi
+      #   run: |
+      #     twine check dist/*
+      #     twine upload --repository-url https://test.pypi.org/legacy/ dist/*
+      #   env:
+      #     TWINE_USERNAME: ${{ secrets.TEST_PYPI_USERNAME }}
+      #     TWINE_PASSWORD: ${{ secrets.TEST_PYPI_PASSWORD }}
```
n-dusan, line 93: Let's remove test pypi code

setup.py
n-dusan, line 1: I believe we want to revert these changes to PACKAGE_NAME and VERSION?

.github/workflows/ci.yml
```diff
@@ -41,21 +44,84 @@ jobs:
...
+      - name: Get Upload URL
+        id: get_upload_url
+        run: echo "${{ github.event.release.upload_url  }}"
+
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.10'
+          # the newest python that we support
+          python-version: '3.12'
```
n-dusan, line 124: Are there any ways to minimize us forgetting to upgrade this variable once newer python versions are out? Could we get the latest python version from the python-version matrix? Or instead create a new variable `latest_python_version`?

n-dusan (changes requested):
Awesome. I was able to install `test_taf` and confirm that taf commands work without any extra installs. Also `taf yubikey export-pub-key` works for both the Yubikey v4 and Yubikey v5.

- As part of this PR, could we also bring the top-level [README.md](https://github.com/openlawlibrary/taf/blob/e95e25fadd238a19df84e48c2fcba6933e30951b/README.md) up to date? Since we won't be using Azure anymore and we aren't building wheels on Ubuntu 16 and 18?

- I see Node deprecation warnings in the GitHub Actions build [here](https://github.com/openlawlibrary/taf/actions/runs/9707362671). Could we resolve? I think it should be really easy, we bump the version numbers on action jobs that are causing issues

.github/workflows/ci.yml
renatav: We typically release TAF by drafting a new release. We also need the release URL to upload the assets

.github/workflows/ci.yml
renatav: Moved this to docs. I'd like to make sure that we don't have to figure out again how to test the release to PyPi.

.github/workflows/ci.yml
renatav: I spent a lot of time trying to get the last element of an array. I couldn't not figure out how to do it, if it's even possible. `-1` does not work. Getting the first element of an array is achievable. Probably spent too much time working on it, since it's not even that important to update this version, as long as the version is still supported. Using a variable is not a problem.

n-dusan (approved):

PR #459, opened 2024-06-29
feat: added bare for cloning and updating

### LINKED ISSUE #413: updater: support updating bare repositories
Make updater work with bare repositories.

Currently if we run `taf repo update` in a bare repo, we get the following error:
```
(oll) λ taf repo update
Updating repository
Updating repository <org>/<law>
Cloning repository and running TUF updater...
Validation repository: cloning repository
Validation repository: trying to clone from https://github.com/<org>/<law>
Validation repository: successfully cloned from https://github.com/<org>/<law>
Running TUF validation of the authentication repository...
WARNING: Could not validate authentication repository at revision 71d9678473ad77d4c24f197bba228481d7900b01 due to error: Final root.json is expired
An error occurred while updating repository cityofsanmateo/law while running step clone_remote_and_run_tuf_updater: D:\OLL\oll-test-repos\delete-me\<org>\<law> is not a Git repository. Run `taf repo clone` instead
Called LifecycleStage.UPDATE handler. Event: failed
WARNING: Config file D:\OLL\oll-test-repos\delete-me\config.json not found. Scripts might not be executed successfully
Update of cityofsanmateo/law failed due to error: D:\OLL\oll-test-repos\delete-me\<org>\<law> is not a Git repository. Run `taf repo clone` instead
```

ronantakizawa:

Added --bare tags for repository cloning and updating to clone and update bare repositories. Fix #413


taf/updater/updater_pipeline.py
```diff
@@ -360,7 +364,7 @@ def clone_remote_and_run_tuf_updater(self):
             # Use ThreadPoolExecutor to run _clone_validation_repo in a separate thread
             with ThreadPoolExecutor() as executor:
                 future_validation_repo = executor.submit(
-                    _clone_validation_repo, self.url
+                    _clone_validation_repo, self.url, self.bare
```
renatav, line 367: See previous comments. This should always be a bare rep

taf/updater/updater_pipeline.py
```diff
@@ -401,29 +406,26 @@ def clone_remote_and_run_tuf_updater(self):
                     )
                 else:
                     self.state.auth_repo_name = auth_repo_name
-
             self.state.users_auth_repo = AuthenticationRepository(
-                self.library_dir,
-                self.state.auth_repo_name,
+                path=Path(self.library_dir) / self.state.auth_repo_name,
```
renatav, line 410: You should not have to do that, `path` is set automatically when you specify `library` and `name`, There's an issue somewhere if that is not the case atm

taf/updater/updater_pipeline.py
```diff
@@ -633,11 +637,11 @@ def clone_repo_to_temp(temp_repo, users_repo):
                     temp_repo.clone_from_disk(
                         users_repo.path,
                         users_repo.get_remote_url(),
-                        is_bare=True,
+                        is_bare=self.bare,
```
renatav, line 640: All repositories which are cloned to temp should always be bare. We do not want to have worktrees before we run validation. So first we clone/move everything to temp, validate, and then clone/update user's repositories. The user's repositories can be bare or not bare.

taf/updater/updater_pipeline.py
```diff
@@ -1294,9 +1302,9 @@ def _clone_validation_repo(url):
     temp_dir = tempfile.mkdtemp()
     path = Path(temp_dir, "auth_repo").absolute()
     validation_auth_repo = AuthenticationRepository(
-        path=path, urls=[url], alias="Validation repository"
+        path=path, urls=[url], alias="Validation repository", bare=bare
     )
-    validation_auth_repo.clone(bare=True)
```
renatav, line 1299: Should still be `True`, see previous comments

renatav:
This update process is really complex, so I tried to explain a few important considerations. Let me know if providing more details would be helpful.

taf/git.py
```diff
@@ -150,15 +150,32 @@ def remotes(self) -> List[str]:
         return self._remotes
 
     @property
-    def is_git_repository_root(self) -> bool:
-        """Check if path is git repository."""
-        git_path = self.path / ".git"
-        return self.is_git_repository and (git_path.is_dir() or git_path.is_file())
+    def is_git_repository(self) -> bool:
+        try:
+            repo = pygit2.Repository(str(self.path))
```
renatav, line 155: It's better to use `self.pygit_repo` repo here because this property handles errors when instantiating pygit2 repository instances

taf/git.py
```diff
@@ -150,15 +150,32 @@ def remotes(self) -> List[str]:
...
+            if self.is_bare_repository and not repo.is_bare:
+                return False
+            if not self.is_bare_repository and repo.head is None:
```
renatav, line 158: I don't think that this check will work if the repository is empty (just initialized)

taf/git.py
```diff
@@ -150,15 +150,32 @@ def remotes(self) -> List[str]:
...
 
     @property
-    def is_git_repository_root(self) -> bool:
-        """Check if path is git repository."""
-        git_path = self.path / ".git"
-        return self.is_git_repository and (git_path.is_dir() or git_path.is_file())
+    def is_git_repository(self) -> bool:
+        try:
+            repo = pygit2.Repository(str(self.path))
+            if self.is_bare_repository and not repo.is_bare:
```
renatav, line 156: I think that `self.is_bare_repository` returns `repo.is_bare`. If `self.is_bare_repository` returns `True`, are we already sure that thar direcotry contains a repository?

taf/git.py
```diff
@@ -150,15 +150,32 @@ def remotes(self) -> List[str]:
...
 
     @property
-    def is_git_repository(self) -> bool:
-        repo_path = pygit2.discover_repository(str(self.path))
-        return repo_path is not None
+    def is_git_repository_root(self) -> bool:
+        if not self.is_git_repository:
+            return False
+        try:
+            repo = pygit2.Repository(str(self.path))
```
renatav, line 169: Same suggestion about not calling `pygit2.Repository` directly

taf/git.py
```diff
@@ -150,15 +150,32 @@ def remotes(self) -> List[str]:
...
+            if self.is_bare_repository:
+                return repo.is_bare and Path(repo.path).resolve() == self.path.resolve()
```
renatav, line 171: I really am not sure that this checks are exactly what we want here. It would be easy to tests if we actually had a test. There is a `test_git.py` test which we just recently added and only contains one tests. Could you add a new test, where you'd initialize a bare repository (using this module's `init` functionality`) then validate these changes? We should also test the case when the repository is not bare

taf/git.py
```diff
@@ -589,8 +615,10 @@ def cleanup(self):
             self._pygit = None
 
     def clone(
-        self, no_checkout: bool = False, bare: Optional[bool] = False, **kwargs
+        self, no_checkout: bool = False, bare: Optional[bool] = None, **kwargs
     ) -> None:
+        if bare is None:
+            bare = self.is_bare_repository
```
renatav, line 621: When cloning, the repository should not already exist, so it makes more sense to pass in `False` then `None` as a default value.

taf/repository_tool.py
```diff
@@ -171,8 +171,9 @@ def _check_key_and_get_pin(expected_key_id):
 
 
 class Repository:
-    def __init__(self, path, name="default"):
+    def __init__(self, path, bare=False, name="default"):
```
renatav, line 174: Probably forgot to remove this `bare`

taf/updater/handlers.py
```diff
@@ -76,13 +76,15 @@ def metadata_dir(self):
...
         return str(self.validation_auth_repo.path / "targets")
 
-    def __init__(self, auth_url, repository_directory, repository_name):
+    def __init__(self, auth_url, repository_directory, repository_name, bare=False):
         """
         Args:
         auth_url: repository url of the git repository which we want to clone.
         repository_directory: the client's local repository's location
         repository_name: name of the repository in 'organization/namespace' format.
+        bare: if True, indicates that the repository is a bare repository.
```
renatav, line 85: Is this still used?

taf/updater/updater.py
```diff
@@ -186,6 +186,10 @@ class RepositoryConfig:
         default=False,
         metadata={"docs": "Whether update fails if a warning is raised. Optional."},
     )
+    bare: bool = field(
+        default=False,
+        metadata={"docs": "Whether the repository is bare. Optional."},
```
renatav, line 191: We should make it clear that all repositories will be cloned as bare

taf/updater/updater.py
```diff
@@ -263,9 +267,6 @@ def update_repository(config: RepositoryConfig):
     """
     settings.strict = config.strict
 
-    # if path is not specified, name should be read from info.json
```
renatav, line 266: This was probably removed by the extension. We should keep this comment

taf/updater/updater.py
```diff
@@ -279,7 +280,18 @@ def update_repository(config: RepositoryConfig):
         if config.url is None:
             raise UpdateFailedError("URL cannot be determined. Please specify it")
 
-    return _update_or_clone_repository(config)
```
renatav, line 282: I don't think that should be necessary this line. `_update_or_clone_repository` performs `validation`, so no need to call both. We also need to keep the `return` here. We collect and return some data that is then used by out lifecycle handlers. They allow the users to get access to the information about the update that was run (was it successful, what was pulled etc.)

taf/updater/updater.py
```diff
@@ -654,6 +673,7 @@ def validate_repository(
             only_validate=True,
             validate_from_commit=validate_from_commit,
             excluded_target_globs=excluded_target_globs,
+            bare=bare,
```
renatav, line 676: `validate` probably does not need this input. It is only called when the repositories are already on disk, in which case we can automatically determine if a repository is a bare repository

taf/updater/updater.py
```diff
@@ -625,6 +639,7 @@ def validate_repository(
     validate_from_commit=None,
     excluded_target_globs=None,
     strict=False,
+    bare=False,
```
renatav, line 642: Validation is only run when the repositories already exist, to validate local repositories. So, in this case, we could check if the repositories are bare.

taf/updater/updater_pipeline.py
```diff
@@ -401,29 +403,25 @@ def clone_remote_and_run_tuf_updater(self):
                     )
                 else:
                     self.state.auth_repo_name = auth_repo_name
-
             self.state.users_auth_repo = AuthenticationRepository(
-                self.library_dir,
```
renatav, line 406: When we instantiate the `AuthenticationRepository`, we can specify `library_dir` and `name`, in which case the path is set by joining those two parameters. If we specify the path, we then determine `library_dir` and `name` based on it. This is fine, we can specify all 3, just noting that it should not be necessary

renatav (commented):
I cloned the repository with the `--bare` flag and that was successful. All repositories are indeed bare repositories! Ran `taf repo update` following the clone, but getting an error:

```
An error occurred while updating authentication repository while running step check_if_local_repositories_clean: Repository mohicanlaw/law has uncommitted changes. Commit and push or revert the changes and run the command again.
```
`check_if_local_repositories_clean` should be updated to handle bare repositories too. Can we even skip it?

taf/updater/updater.py
ronantakizawa: Are you talking about adding bare as a parameter to the validate_repository function? If so I'd have taf repo validate check if the auth_repo is a bare repo.

taf/git.py
```diff
@@ -150,15 +150,38 @@ def remotes(self) -> List[str]:
...
-        return self.is_git_repository and (git_path.is_dir() or git_path.is_file())
+    def is_git_repository(self) -> bool:
+        try:
+            repo = self.pygit_repo
+            if repo is None:
+                return False
+            # Check if the repository is bare
+            if self.is_bare_repository:
+                return repo.is_bare
+            # Check if the repository has a HEAD
```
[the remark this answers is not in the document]
ronantakizawa: I get errors using pygit2.discover_repository(str(self.path)) in is_git_repository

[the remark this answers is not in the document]
ronantakizawa: Nvm got it to work.

renatav (approved):
This is proving to be a more complex issue than I initially anticipated. This does work now and unauthenticated commits are removed correctly. However, when we run update, we always get this message:
`Repository x/y: states of target repositories are not in sync with last validated commit. Starting the update from the beginning`
What seems to be happening is the following:
- before updating target repositories, we clone them to temp (basically, copy from the user's filesystem, we do not clone them from GitHub again). This is done in the `clone_target_repositories_to_temp` step
- in `determine_start_commits` we check if the states of target repositories are consistent with the authentication repository (users can manually pull without running the updater for example). If that is not the case, we validate from the beginning (this is just some workaround, we want to handle that more efficiently). 
- In this case, with bare repositories, these repositories which are in temp do not have all local branches (remote branches are all there) that we expect. This is causing the update to start from the beginning. This is where that is in the code:
```
                    current_branch = last_validated_repository_commits_data.get(
                            "branch", repository.default_branch
                        )
                        last_validated_commit = last_validated_repository_commits_data[
                            "commit"
                        ]
                        branch_exists = repository.branch_exists(
                            current_branch, include_remotes=False
                        )
                        if not branch_exists:
                            is_initial_state_in_sync = False
                            break
                        top_commit_of_branch = repository.top_commit_of_branch(
                            current_branch
                        )
                        if top_commit_of_branch != last_validated_commit:
```
We need to figure out how to handle that.

I'm going to approve this pull request, because this technically works and there are no regressions as far as I can tell, so feel free to merge if you want to. I can wrap that up, I don't think that debugging this is a good use of your time. If you want to replicate the issue and give it a stab, just ran `taf repo update` after cloning with `--bare` and notice that "starting update from the beginning message"

taf/git.py
```diff
@@ -1623,6 +1623,30 @@ def _determine_default_branch(self) -> Optional[str]:
...
 
+    def top_commit_of_remote_branch(
+        self, branch, remote="origin", users_target_repositories=None
+    ):
+        """
+        Fetches the top commit of the specified remote branch.
+        If the remote branch does not exist, checks if the local repository has the branch and creates it if necessary.
+        """
+        remote_branch = f"{remote}/{branch}"
+        if not self.branch_exists(remote_branch, include_remotes=True):
```
renatav, line 1634: `git.py` shouldn't be aware of the existence of other repositories, it's responsibilities should be strictly related to executing git commands. Sorry if my comment was unclear. This should just be:
```
 remote_branch = f"{remote}/{branch}" 
return self.top_commit_of_branch(remote_branch)
```
Everything else should still be in `updater_pipeline`

taf/updater/updater_pipeline.py
```diff
@@ -802,23 +802,22 @@ def determine_start_commits(self):
...
+                            is_initial_state_in_sync = False
+                            break
                         if repository.is_bare_repository:
-                            # fetches the top commit of the remote branch (origin/{current_branch})
-                            remote_branch = f"origin/{current_branch}"
-                            top_commit_of_branch = repository.top_commit_of_branch(
-                                remote_branch
-                            )
+                            try:
+                                top_commit_of_branch = repository.top_commit_of_remote_branch(
```
renatav, line 813: All of the checks that we have for regular repos should still be executed even if this is a bare repo. The only difference is that, since after being clone from disk to temp local branches do not seem to get created, checking if the branch exists and getting the top commit will need to done differently. Once we've checked if the branch exists and found the top commit, we still need:
```
    if top_commit_of_branch != last_validated_commit:
                            # check if top commit is newer (which is fine, it will be validated)
                            # or older, meaning that the authentication repository contains
                            # additional commits, so it would be necessary to find older auth repo
                            # commit and start the validation from there
                            if (
                                current_branch
                                not in repository.branches_containing_commit(
                                    last_validated_commit
                                )
                            ):
                                is_initial_state_in_sync = False
                                break
```
I'm worried that `branches_containing_commit` will not work if we don't have the local branches.

I've been looking at this and I can't remember why I even decided to iterate over temp repositories here. We might be able to get rid of all of this code (since local branches should exist by iterating over local repositories here `self.state.users_target_repositories`. Let's try doing that, and if there are any other issues, I'll debug tomorrow. We can keep the new git method once it's cleaned up in any case, it might be useful

renatav (approved):
It might be necessary to revisit this as a part of the handling different states issues, but I think that it makes sense to merge now. I tried to test by removing the top commit of the auth repo, which was only the update timestamp and snapshot expiration date commit, revert the last validated commit and run update, and it had to be started from the beginning. Might make more sense to address/test that later than to keep working on this PR, the issue might not even be specific to bare repos

PR #460, opened 2024-07-02
feat: added git hook check for updater

ronantakizawa:

Added git hook check to updater. Resolves issue #393

### DIFF
--- a/taf/updater/updater_pipeline.py
+++ b/taf/updater/updater_pipeline.py
@@ -25,7 +25,7 @@ from taf.exceptions import (
 from taf.updater.handlers import GitUpdater
 from taf.updater.lifecycle_handlers import Event
 from taf.updater.types.update import OperationType, UpdateType
-from taf.utils import TempPartition, on_rm_error
+from taf.utils import TempPartition, on_rm_error, set_executable_permission
 from taf.log import taf_logger
 from tuf.ngclient.updater import Updater
 from tuf.repository_tool import TARGETS_DIRECTORY_NAME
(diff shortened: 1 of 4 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -1215,6 +1216,7 @@ def set_target_repositories_data(self):
 
                 targets_data[repo_name]["commits"] = branch_data
 
+            ensure_pre_push_hook(self.state.users_auth_repo.path)
```
renatav, line 1219: I think that this should be a separate step. Because the updater is confusing enough as it is, I'd like all of the steps to be descriptive and as small as possible. So let's add `ensure_pre_push_hook` to the `steps` array. I think that that should be the last step, so that the message that the hook was added is visible without scrolling.

taf/updater/updater_pipeline.py
```diff
@@ -1512,3 +1514,14 @@ def _merge_commit(repository, branch, commit_to_merge, force_revert=True):
             format_commit(commit_to_merge),
             branch,
         )
+
+
+def ensure_pre_push_hook(auth_repo_path):
```
renatav, line 1519: If we already have this code elsewhere, can we reuse that one? Are there any differences between the two?

renatav (commented):
It works, just some thoughts on how to make the code a bit nicer

taf/utils.py
```diff
@@ -388,6 +367,42 @@ def get_file_details(
...
+        # Check if permissions were set correctly on Unix-like systems
+        if platform.system() != "Windows" and not os.access(pre_push_script, os.X_OK):
+            taf_logger.error(
+                f"Failed to set pre-push git hook executable permission. Please set it manually for {pre_push_script}."
+            )
+            return False
+        taf_logger.info("Pre-push hook not present. Pre-push hook added successfully.")
+        return True
+
+    return False
```
renatav, line 403: I think that we should return `True` if the hook already exists, since `False` implies that an error occurred

renatav (commented):
Works and looks good! Please see one minor comment and update changelog

renatav (approved):

PR #466, opened 2024-07-05
feat: improved usability (TAF finds the repo if current directory has no repo, create a .taf directory to manage keys)

### LINKED ISSUE #461: Usability improvements
Users often get errors when calling commands which need the authentication repository because they invoked them outside of the authentication repository. By default, auth repo path is set to the current working directory. Although this location can be set by using the `--path` parameter, most users rarely use it and prefer to invoke the command from the correct location. These errors can look like: 
```
taf.exceptions.ValidationFailedError: Validation of repository failed due to error: Cannot clone tmpfv5f0916/auth_repo from any of the following URLs: ['.../library/cityofsanmateo']
```
In order to enhance the tool's usability, the following should be done:
- [ ] Try to find the authentication repository based on the current working directory.
  - [ ] when inside the directory which contains just one auth repository, return the location of that auth repo
  - [ ] if inside the auth repo, say, `auth/targets`, return that repository's root
-   [ ] if that cannot be done, raise a nice error message, saying that the auth repository could not be located

The second issue is related to key management, since the users constantly have to specify keystore location.

- [ ] Add a command for initializing a `.taf` directory inside a directory of choice (this should be the library root). Inside that directory, create `config.toml` (can be empty for now) and a keystore directory. Optionally generate new keys
- [ ] If the library root of the current authentication repository contains this `.taf/keystore` directory, read the keystore files from there. Keep the option to specify a different location.

Extend with all other tasks related to usability...

ronantakizawa:

 Improved usability (TAF finds the repo if current directory has no repo, create a .taf directory to manage keys). 
Fix #461 

### DIFF
--- a/taf/api/conf.py
+++ b/taf/api/conf.py
@@ -4,20 +4,6 @@ from pathlib import Path
 from taf.api.keystore import generate_keys
 from taf.log import taf_logger
 
-def find_taf_directory():
-    """Look for the .taf directory within the library root."""
-    library_root = (
-        Path(__file__).resolve().parent.parent
-    )  # Adjusted to determine the library root
-    print(library_root)
-    current_dir = library_root
-    while current_dir != current_dir.root:
-        taf_directory = current_dir / ".taf"
-        if taf_directory.exists() and taf_directory.is_dir():
-            return taf_directory
-        current_dir = current_dir.parent
-    return None
-
 def init(path: Optional[str] = None, build_keys: bool = False):
     # Determine the directory path
     if path:
--- a/taf/api/conf.py
+++ b/taf/api/conf.py
@@ -1,4 +1,4 @@
-from shutil import copytree
+from shutil import Error, copytree
 from typing import Optional
 from pathlib import Path
 from taf.api.keystore import generate_keys
@@ -7,7 +7,6 @@ from taf.log import taf_logger
 
 def init(
     path: Optional[str] = None,
-    should_generate_keys: bool = False,
     keystore: Optional[str] = None,
     roles_key_infos: Optional[str] = None,
 ):
@@ -33,7 +32,7 @@ def init(
     keystore_directory.mkdir(exist_ok=True)
 
     # If any of these parameters exist you can assume the user wants to generate keys
-    if not should_generate_keys and not keystore and not roles_key_infos:
+    if not keystore and not roles_key_infos:
         # Prompt the user if they want to run the generate_keys function
         while True:
             use_keystore = (
@@ -69,8 +68,16 @@ def init(
                         )
         # Check if keystore is specified now. If so copy the keys
         if keystore:
-            copytree(keystore, keystore_directory, dirs_exist_ok=True)
-            taf_logger.info(f"Copied keystore from {keystore} to {keystore_directory}")
+            try:
+                copytree(keystore, keystore_directory, dirs_exist_ok=True)
+                taf_logger.info(
+                    f"Copied keystore from {keystore} to {keystore_directory}"
+                )
+            except FileNotFoundError:
+                taf_logger.error(f"Provided keystore path {keystore} not found.")
+            except Error as e:
+                taf_logger.error(f"Error occurred while copying keystore: {e}")
+
         # If there is no keystore path specified, ask for keys description and generate keys
         elif not roles_key_infos:
             roles_key_infos = input(
--- a/taf/api/conf.py
+++ b/taf/api/conf.py
@@ -1,0 +1,64 @@
+from logging import INFO
+from typing import Optional
+from pathlib import Path
+from taf.api.keystore import generate_keys
+from taf.log import taf_logger
+
+def find_taf_directory():
+    """Look for the .taf directory within the library root."""
+    library_root = (
+        Path(__file__).resolve().parent.parent
+    )  # Adjusted to determine the library root
+    print(library_root)
+    current_dir = library_root
+    while current_dir != current_dir.root:
+        taf_directory = current_dir / ".taf"
+        if taf_directory.exists() and taf_directory.is_dir():
+            return taf_directory
+        current_dir = current_dir.parent
+    return None
+
+def init(path: Optional[str] = None, build_keys: bool = False):
+    # Determine the directory path
+    if path:
+        taf_directory = Path(path) / ".taf"
+    else:
+        taf_directory = Path(".taf")
+
+    if taf_directory.exists() and taf_directory.is_dir():
+        taf_logger.info(".taf directory already exists.")
+    else:
+        # Create the .taf directory
+        taf_directory.mkdir(exist_ok=True)
+
+    # Create the config.toml file
+    config_file_path = taf_directory / "config.toml"
+    config_file_path.touch()  # Create an empty file
+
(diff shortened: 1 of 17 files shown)


taf/api/keystore.py
```diff
@@ -85,3 +85,40 @@ def generate_keys(keystore: Optional[str], roles_key_infos: str) -> None:
...
+    # Create the .taf directory
+    taf_directory = Path(".taf")
+    taf_directory.mkdir(exist_ok=True)
+
+    # Create the config.toml file
+    config_file_path = taf_directory / "config.toml"
+    config_file_path.touch()  # Create an empty file
+
+    # Create the keystore directory
+    keystore_directory = taf_directory / "keystore"
```
renatav, line 101: Haven't tested in detail, but one thing that should also be done is to set the keystore path to this location if the keystore path is not specified (this should be the case for all commands where the keystore is used). That means that it will be necessary to look for this `.taf` directory. We can expect to find it inside the library root.

taf/api/keystore.py
```diff
@@ -66,6 +81,12 @@ def generate_keys(keystore: Optional[str], roles_key_infos: str) -> None:
     Returns:
         None
     """
+    if keystore is None:
```
renatav, line 84: There is a `load_signing_keys` function, where we attempt to load keys from keystore. We should extend that function too and try to find the default keystore location if keystore location is not specified. There is also `setup_roles_keys`. We could attempt to find the default keystore location there as well

taf/utils.py
```diff
@@ -403,6 +403,19 @@ def ensure_pre_push_hook(auth_repo_path: Path) -> bool:
     return True
 
 
+def find_keystore(path: Path) -> Optional[Path]:
+    """
+    Find keystore starting from the given path and traversing parent directories if needed.
```
renatav, line 408: I think that this function should simply call `find_taf_directory` and then check if there is a keystore directory directly inside it.

renatav (commented):
Managed to call `taf repo update` from various locations, which is great. Mostly minor comments, this looks pretty good

taf/tools/conf/__init__.py
```diff
@@ -0,0 +1,30 @@
+import click
+from taf.api.conf import init
+
+
+def init_command():
+    @click.option("--path", default=".", help=".taf directory's location. If not specified, set to the current directory")
+    @click.option("--should-generate-keys", is_flag=True, default=False, help="Specify to generate keys after creating .taf directory.")
+    @click.option("--keystore", default=None, help="Location of the keystore directory. Can be specified in keys-description dictionary")
```
renatav, line 8: Nitpick: please note that this where the keys will be copied from.

taf/api/utils/_conf.py
```diff
@@ -0,0 +1,39 @@
...
+
+    taf_logger.info(f"No .taf directory found starting from {archive_root}")
+    return None
+
+
+def find_keystore(path: Path) -> Optional[Path]:
+    """
+    Find keystore starting from the given path and traversing parent directories if needed.
+    """
+    for parent in [path] + list(path.parents):
```
renatav, line 33: Tested and it does not seem like this was able to find the keystore. It's because it's in one of the parent directories / .taf. Tested
```
    path = find_taf_directory(path) / "keystore"
    if path.is_dir():
        return path
``` 
And that worked

taf/tools/roles/__init__.py
```diff
@@ -25,9 +26,7 @@ def add_role_command():
     @click.option("--no-commit", is_flag=True, default=False, help="Indicates that the changes should not be committed automatically")
     @click.option("--prompt-for-keys", is_flag=True, default=False, help="Whether to ask the user to enter their key if not located inside the keystore directory")
     def add(role, path, parent_role, delegated_path, keystore, keys_number, threshold, yubikey, scheme, no_commit, prompt_for_keys):
-        if not path:
-            print("Specify at least one path")
```
renatav, line 29: Could you keep this and check if `delegated_path` is None istead? Let's fix that since we are already modifying this code.

renatav (commented):
Looks good, almost ready to be merged.

taf/api/conf.py
```diff
@@ -0,0 +1,85 @@
...
+                    if keystore_path.exists() and keystore_path.is_dir():
+                        keystore = keystore_input  # Assign the string path to the keystore variable
+                        break
+                    else:
+                        taf_logger.error(
+                            f"Provided keystore path {keystore} is invalid."
+                        )
+        # Check if keystore is specified now. If so copy the keys
+        if keystore:
+            copytree(keystore, keystore_directory, dirs_exist_ok=True)
```
renatav, line 72: This will raise an error if the path is not correct: 
``` 
in copytree
    with os.scandir(src) as itr:
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'keystore-dev2334'
```
Let's just print a nice error message.

taf/api/conf.py
```diff
@@ -0,0 +1,85 @@
+from shutil import copytree
+from typing import Optional
+from pathlib import Path
+from taf.api.keystore import generate_keys
+from taf.log import taf_logger
+
+
+def init(
+    path: Optional[str] = None,
+    should_generate_keys: bool = False,
```
renatav, line 10: Now that I've had a chance to run this command a few times, I think that we don't even need this. If none of `keystore` location (location from which we want to copy the keys) and `roles_key_infos` are specified, we could ask them if they want to generate the keys. When generating new keys, the users will be prompted to enter a couple of things, so not having to press one more `y`  does not make a different. If you decide to keep it, I would name the CLI option `generate_keys`, but this parameter can still be `should_generate_keys`. Or use an alias when importing `generate_keys`

taf/api/utils/_conf.py
```diff
@@ -0,0 +1,38 @@
...
+    """Look for the .taf directory within the archive root.
+
+    Args:
+        auth_repo_path (Path): The path to the authentication repository.
+
+    Returns:
+        Optional[Path]: The path to the .taf directory if found, otherwise None.
+    """
+    # Determine the archive root
+    archive_root = auth_repo_path.parent.parent
```
renatav, line 16: Let's start from `auth_repo_path.parent`, just in case

taf/keys.py
```diff
@@ -236,11 +237,14 @@ def load_signing_keys(
     # if the keystore file is not found, ask the user if they want to sign
     # using yubikey and to insert it if that is the case
 
-    keystore_path = None
-    if keystore is not None:
-        keystore_path = Path(keystore).expanduser().resolve()
-    else:
-        taf_logger.info("Keystore location not provided")
+    if keystore is None:
+        keystore_path = find_keystore(Path.cwd())
```
renatav, line 241: `find_taf_directory` expects `auth_repo_path`, so this couldn't find the keystore when I was in the auth repo's parent directory. Everything worked when I was in `D:\oll\library\oll-test-repos\cityofsanmateo\law`, but not when I was in ``D:\oll\library\oll-test-repos\cityofsanmateo` (the repository was found, keystore was not)

renatav (commented):
This is looking great. I found some minor issues while testing

renatav (approved):

PR #471, opened 2024-07-09
feat: added --force to taf repo update

### LINKED ISSUE #467: Add --force flag to the updater. Clean up local repositories before running update
In certain cases, update fails or is restarted because repositories are in inconsistent state. This can happen because the user can update their local repositories without the updater (by manually pulling, committing to these repos, or even removing some commits). In the past, the user often had to remove the repositories and clone again. While we want to keep improving the updater and making it more robust, we also want to improve the user experience in cases when there is a risk that the update will fail.

Add `--force` flag to `taf repo update`. It should be set to `False` by default. This flag should have the following effect on the update process:
- if repositories are not clean,  run `git clean -fd; git reset HEAD --hard`, proceed with update
  - this can be added the `check_if_local_repositories_clean` step
- if the authentication repository's top commit is the same as the last validated commit, but target repositories are not on commits specified in authentication repository, reset target repositories to those commits. These commits are often referred to as "target commits" in the code and are stored in target files, see https://github.com/cityofsanmateo/law/blob/master/targets/cityofsanmateo/law-xml
  - a lot of the code that is needed to implement this is already in `determine_start_commits`. In fact, determine start commits can be extended to support this - reset instead of noting that some of the repositories are out of sync
- if there are unpushed commits, remove them, sync the repositories with last validated commits

## Local repositories not in sync with last validated commit

After each successful update, the updater writes the top commit of the auth repo to a file called `last_validated_commit`. Without any manual updates, top commit of the auth repo will always be the same as the commit stored in this file. However, the user can manually pull or even run `git reset --hard HEAD~x` and remove a number of valid commits. Similarly, target repositories can be in sync with the auth repository, which means that commit shas stored in the most recent auth repo target files are the same as the top commits of these repositories. However, a user can also manually pull new changes or remove valid commits.
- if auth repo's top commit is not the same as the last validated commit, check if if last validated commit is auth repo's history (I think that this is already the case)
  - if it is, update/validation can still be run. Updater will either confirm that these are valid commits (manually pulled), and successfully update the repositories, or fail.
  - if that's not the case, that could mean that some comments were removed. Currently, validation will be run from the beginning and if it is successful, everything will be in sync. This will required `--force=upstream`
- Same thing with target repositroies

Add the following tests:
 - all of these cases where last validated commits are not the same as top commits of repositories(with and without the `--force` flag). Some of them might already exist
 - update when one or more target repositories are in a detached head state

ronantakizawa:

Added forced updates to the updater. Fix #467.

### DIFF
--- a/taf/<taf.auth_repo.AuthenticationRepository object at 0x1076801d0>/dirty_file.txt
+++ b/taf/<taf.auth_repo.AuthenticationRepository object at 0x1076801d0>/dirty_file.txt
@@ -1,0 +1,1 @@
+GGqLfFacTc
--- a/taf/<taf.auth_repo.AuthenticationRepository object at 0x1076801d0>/dirty_file.txt
+++ b/taf/<taf.auth_repo.AuthenticationRepository object at 0x1076801d0>/dirty_file.txt
@@ -1,1 +1,0 @@
-GGqLfFacTc
--- a/taf/git.py
+++ b/taf/git.py
@@ -995,6 +995,16 @@ class GitRepository:
         """Get commit sha of HEAD~{behind_head}"""
         return self._git("rev-parse HEAD~{}", behind_head)
 
+    def get_current_commit(self) -> Optional[str]:
+        """
+        Returns the SHA of the current commit.
+        """
+        try:
+            return self._git("rev-parse HEAD").strip()
+        except GitError as e:
+            self._log_error(f"Error getting current commit: {e}")
+            return None
+
     def get_default_branch(self, url: Optional[str] = None) -> str:
         """Get the default branch of the repository. If url is provided, return the
         default branch from the remote. Otherwise, return the default
--- a/taf/git.py
+++ b/taf/git.py
@@ -642,6 +642,14 @@ class GitRepository:
             self._pygit.cleanup()
             self._pygit = None
 
+    def clean_and_reset(self):
+        """Cleans the untracked files and resets the HEAD to the latest commit."""
+        try:
+            self.clean()
+            self.reset_to_head()
+        except GitError as e:
+            raise GitError(f"Failed to clean and reset the repository: {e}")
+
     def clone(
         self, no_checkout: bool = False, bare: Optional[bool] = False, **kwargs
     ) -> None:
--- a/taf/git.py
+++ b/taf/git.py
@@ -995,16 +995,6 @@ class GitRepository:
         """Get commit sha of HEAD~{behind_head}"""
         return self._git("rev-parse HEAD~{}", behind_head)
 
-    def get_current_commit(self) -> Optional[str]:
-        """
-        Returns the SHA of the current commit.
-        """
-        try:
-            return self._git("rev-parse HEAD").strip()
-        except GitError as e:
-            self._log_error(f"Error getting current commit: {e}")
-            return None
-
     def get_default_branch(self, url: Optional[str] = None) -> str:
         """Get the default branch of the repository. If url is provided, return the
         default branch from the remote. Otherwise, return the default
--- a/taf/git.py
+++ b/taf/git.py
@@ -648,7 +648,9 @@ class GitRepository:
             self.clean()
             self.reset_to_head()
         except GitError as e:
-            raise GitError(f"Failed to clean and reset the repository: {e}")
+            raise GitError(
+                self, message=f"Failed to clean and reset the repository: {e}"
+            )
 
     def clone(
         self, no_checkout: bool = False, bare: Optional[bool] = False, **kwargs
--- a/taf/tests/test_updater/conftest.py
+++ b/taf/tests/test_updater/conftest.py
@@ -1,4 +1,5 @@
 import enum
+import os
 import re
 import pytest
 import inspect
@@ -18,6 +19,7 @@ from taf.auth_repo import AuthenticationRepository
 from taf.constants import DEFAULT_RSA_SIGNATURE_SCHEME
 from taf.messages import git_commit_message
 from taf import repositoriesdb, settings
+from taf.exceptions import GitError
 from taf.utils import on_rm_error
 from taf.log import disable_console_logging
 from taf.tests.test_updater.update_utils import load_target_repositories
@@ -231,7 +233,14 @@ def clone_client_repo(target_name: str, origin_dir: Path, client_dir: Path):
 def create_repositories_json(library_dir: Path, repo_name: str, targets_config: list):
     repo_path = Path(library_dir, repo_name)
     targets_dir_path = repo_path / TARGETS_DIRECTORY_NAME
-    targets_dir_path.mkdir(parents=True, exist_ok=True)
+
+    try:
+        targets_dir_path.mkdir(parents=True, exist_ok=True)
+    except PermissionError:
+        # Attempt to fix permissions
+        os.chmod(repo_path, 0o775)
+        targets_dir_path.mkdir(parents=True, exist_ok=True)
+
     if len(targets_config):
         repositories_json = generate_repositories_json(targets_config)
         (targets_dir_path / REPOSITORIES_JSON_NAME).write_text(repositories_json)
@@ -602,3 +611,31 @@ def update_target_files(target_repo: GitRepository, commit_message: str):
             new_content = existing_content + "\n" + text_to_add
             file_path.write_text(new_content, encoding="utf-8")
     target_repo.commit(commit_message)
+
+
+def update_file_without_commit(repo_path: str, filename: str, content: str):
+    file_path = Path(repo_path) / filename
+    file_path.parent.mkdir(parents=True, exist_ok=True)  # Ensure the directory exists
+    if file_path.exists():
+        with file_path.open("a") as file:
+            file.write(content)
+    else:
+        with file_path.open("w") as file:
+            file.write(content)
+
+
+def add_file_without_commit(repo_path: str, filename: str, content: str):
+    file_path = Path(repo_path) / filename
+    file_path.parent.mkdir(parents=True, exist_ok=True)  # Ensure the directory exists
+    with file_path.open("w") as file:
+        file.write(content)
+
+
+def remove_commits(repo_path: str, num_of_commits: int, repo_name: str):
+    repo = GitRepository(path=Path(repo_path))
+    print(repo)  # Log the state of the repository object
+
+    try:
+        repo.clean_and_reset()
+    except GitError as e:
+        print(f"{str(e)}")
--- a/taf/tests/test_updater/conftest.py
+++ b/taf/tests/test_updater/conftest.py
@@ -60,6 +60,8 @@ METADATA_EXPIRED = r"Update of (\w+\/\w+) failed due to error: Validation of aut
 NO_INFO_JSON = "Update of repository failed due to error: Error during info.json parse. If the authentication repository's path is not specified, info.json metadata is expected to be in targets/protected"
 UNCOIMITTED_CHANGES = r"Update of (\w+\/\w+) failed due to error: Repository (\w+\/\w+) should contain only committed changes\. \nPlease update the repository at (.+) manually and try again\."
 UPDATE_ERROR_PATTERN = r"Update of (\w+\/\w+) failed due to error: Validation of authentication repository (\w+\/\w+) failed at revision ([0-9a-f]+) due to error: .*"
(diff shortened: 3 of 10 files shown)


renatav (commented):
I haven't taken a detailed look, but this is definitely something that we should cover with tests, which should be pretty easy to do with this new testing framework.

The first thing would be to add tests where we intentionally create invalid states and call the updater without the `--force` flag. That could include the following (we can add more tests later):
- index of auth repo is dirty (one file updated and a new file added, two separate cases)
- index of a target repo is dirty (one file updated and a new file added, two separate cases)
- target repo contains an additional commit (this one already exists)
- one or more commits were removed from the target repo

Add these tests to `test_updater_invalud`.

I don't think that we have a setup functions which updates files without committing, so it will be necessary to add them. Take a look at `    setup_manager.add_task(add_unauthenticated_commits_to_all_target_repos)` for example.

These tests are set up by creating an initial set of repositories and then defining how they should be updated by writing these update functions, like `add_unauthenticated_commits_to_all_target_repos`. Create additional setup functions like that one, but do not commit anything.
Let me know if you'd like me to provide more details about the testing framework.

Then, copy-paste the same tests to `test_update_valid`, but call with the force flag. All tests should pass.

taf/updater/updater_pipeline.py
```diff
@@ -399,14 +401,25 @@ def check_if_local_repositories_clean(self):
...
                 else:
                     if auth_repo.something_to_commit():
-                        raise RepositoryNotCleanError(auth_repo.name)
+                        if self.force:
+                            taf_logger.info(
+                                f"Cleaning and resetting repository {auth_repo.name} for forced update"
+                            )
+                            auth_repo.clean_and_reset()
+                        else:
+                            raise RepositoryNotCleanError(auth_repo.name)
```
renatav, line 410: Let's extend these exceptions, so that if they are by the updater, we print a message saying that the updater can be run with the `--force` flag.

taf/updater/updater_pipeline.py
```diff
@@ -399,14 +401,25 @@ def check_if_local_repositories_clean(self):
...
                     ):
-                        raise UnpushedCommitsError(
-                            auth_repo.name,
-                            auth_repo.default_branch,
-                        )
+                        if self.force:
+                            taf_logger.info(
+                                f"Cleaning and resetting repository {auth_repo.name} for forced update"
+                            )
+                            auth_repo.clean_and_reset()
```
renatav, line 418: If there are unpushed commits, with the `--force` flag, the update is successful, but the commit is not removed. The update probably would've failed if there were fetched commits (merge would fail). We should remove these unpushed commits. We have a couple of other cases here (top commits can be different than the last validated commit). We'll need to handle them as well, but I'll update the issue first. Let's focus on these simpler cases for now. The issue here is that `auth_repo.clean_and_reset()` doesn't remove these additional commits. Let's also make sure that we cover this with a test. After the update, in addition to checking if it was successful, check if the top commit of the repository is equal to the last pushed commit

taf/updater/updater_pipeline.py
```diff
@@ -796,90 +819,95 @@ def determine_start_commits(self):
...
+                last_validated_commit
+            ):
+                return False
+            self.state.old_heads_per_target_repos_branches[repository.name][
+                current_branch
+            ] = last_validated_commit
+
+        return True
+
+    def _reset_target_repositories(self):
```
renatav, line 882: I don't think that we need to call `clean_and_reset()` again. The idea behind this code was to update the state, without making any changes to the repositories. Basically, to make sure that we'll validate everything from the beginning. `clean_and_reset` should get called only if we detect that there a repository contains uncommitted/unpushed changes. You are right though, with this `--force` flag, we will need to make some additional changes if repositories are not in sync. I'll update the issue with the next steps

taf/updater/updater_pipeline.py
```diff
@@ -399,14 +401,38 @@ def check_if_local_repositories_clean(self):
...
-                            auth_repo.default_branch,
-                        )
+                        if self.force:
+                            taf_logger.info(
+                                f"Resetting repository {auth_repo.name} to clean state for a forced update."
+                            )
+                            auth_repo.clean_and_reset()
+                            auth_repo.reset_num_of_commits(
+                                num_of_commits=len(
+                                    auth_repo.all_commits_since_commit(
```
renatav, line 424: Would `reset_to_commit` work here?

renatav (commented):
There is some room to improve these tests, but I'm excited to see progress being made here! I realized that part of this work was already being handled, I apologize for the confusion there (I forgot that I had already implemented a part of this). I renamed the issues, since the next step will be to figure out if we can add a sync with upstream. That will be a separate issue.

taf/git.py
```diff
@@ -995,6 +995,16 @@ def get_commit_sha(self, behind_head: str) -> str:
         """Get commit sha of HEAD~{behind_head}"""
         return self._git("rev-parse HEAD~{}", behind_head)
 
+    def get_current_commit(self) -> Optional[str]:
```
renatav, line 998: Would `head_commit_sha()` work?

taf/<taf.auth_repo.AuthenticationRepository object at 0x1076801d0>/dirty_file.txt
```diff
@@ -0,0 +1 @@
+GGqLfFacTc
```
renatav, line 1: Probably committed accidentally

taf/tests/test_updater/conftest.py
```diff
@@ -602,3 +611,53 @@ def update_target_files(target_repo: GitRepository, commit_message: str):
...
+
+
+def create_index_lock_in_repo(repo_path: str):
+    """Creates an index.lock file to simulate an interrupted git operation."""
+    repo = GitRepository(path=Path(repo_path))
+    index_lock_path = repo.path / ".git" / "index.lock"
+    index_lock_path.touch()
+
+
+def commit_file_to_repo(repo_path: str, file_name: str, commit_message: str):
```
renatav, line 660: If you only need to commit one of the updated files, without previously adding everything that was changed to the index, you'll need to update `git.py` and `commit`. It could have optional input argument `filenames`. If they are specified, add those files to index, if not, add everything (this is how `commit` currently works

taf/tests/test_updater/test_update/test_update_invalid.py
```diff
@@ -73,3 +79,186 @@ def test_update_invalid_repo_target_in_indeterminate_state(
...
+        origin_auth_repo,
+        client_dir,
+    )
+
+    clients_auth_repo_path = client_dir / origin_auth_repo.name
+    update_file_without_commit(str(clients_auth_repo_path), "dirty_file.txt")
+
+    try:
+        # Attempt to update, which should fail due to uncommitted changes
+        update_and_check_commit_shas(
```
renatav, line 104: This looks good now, the update should fail. Instead of calling this with `try-except`, we want to call `update_invalid_repos_and_check_if_repos_exist` because you can specify the error that we expect to be raised.  E.g.
```
update_invalid_repos_and_check_if_repos_exist(
        OperationType.UPDATE,
        origin_auth_repo,
        client_dir,
        TARGET_MISSMATCH_PATTERN,
        True,
    )
``` 
Here we expect `TARGET_MISSMATCH_PATTERN = r"Update of (\w+)\/(\w+) failed due to error: Failure to validate (\w+)\/(\w+) commit ([0-9a-f]{40}) committed on (\d{4}-\d{2}-\d{2}): data repository ([\w\/-]+) was supposed to be at commit ([0-9a-z]{40}) but repo was at ([0-9a-f]{40})"`
This is this function includes:

```
    def _update_expect_error():
        with pytest.raises(UpdateFailedError, match=expected_error):
            if operation == OperationType.CLONE:
                clone_repository(config)
            else:
                update_repository(config)
```
So, we want to use `with pytest.raises`
Let me know if that function does not do the trick

taf/tests/test_updater/test_update/test_update_invalid.py
```diff
@@ -73,3 +79,186 @@ def test_update_invalid_repo_target_in_indeterminate_state(
...
+    indirect=True,
+)
+def test_dirty_index_auth_repo_add_file(origin_auth_repo, client_dir):
+    clone_repositories(
+        origin_auth_repo,
+        client_dir,
+    )
+
+    auth_repo_path = origin_auth_repo.path / "namespace/auth_repo"
+    add_file_without_commit(str(auth_repo_path), "new_file.txt")
```
renatav, line 136: Should be updated like the previous one - update client's repo and specify which error should be raised

taf/tests/test_updater/test_update/test_update_invalid.py
```diff
@@ -73,3 +79,186 @@ def test_update_invalid_repo_target_in_indeterminate_state(
...
+def test_dirty_index_target_repo_add_file(origin_auth_repo, client_dir):
+    clone_repositories(
+        origin_auth_repo,
+        client_dir,
+    )
+
+    target_repo_path = origin_auth_repo.path / "namespace/target1"
+    add_file_without_commit(str(target_repo_path), "new_file.txt")
+
+    update_and_check_commit_shas(
```
renatav, line 188: Same here and other tests where we expect an error to be raised

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -400,3 +404,230 @@ def test_update_valid_when_no_upstream_when_contain_unsigned_commits(
...
+    indirect=True,
+)
+def test_update_valid_dirty_index_auth_repo_add_file(origin_auth_repo, client_dir):
+    clone_repositories(
+        origin_auth_repo,
+        client_dir,
+    )
+
+    auth_repo_path = origin_auth_repo.path / "namespace/auth_repo"
+    add_file_without_commit(str(auth_repo_path), "new_file.txt")
```
renatav, line 451: Make sure that you are modifying client's repo everywhere where we want to test if the update will be successful if client's repo is in some invalid state when called with `--force` flag

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -400,3 +404,230 @@ def test_update_valid_when_no_upstream_when_contain_unsigned_commits(
...
+    )
+
+    # Step 1: Add valid signed target commits
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.execute_tasks()
+
+    # Set last successful commit
+    last_valid_commit = origin_auth_repo.get_current_commit()
+    origin_auth_repo.set_last_validated_commit(last_valid_commit)
```
renatav, line 617: Could we create a setup manager task which does this?

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -400,3 +404,230 @@ def test_update_valid_when_no_upstream_when_contain_unsigned_commits(
...
+
+    # Step 2: Add more valid signed target commits
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.execute_tasks()
+
+    # Step 3: Add invalid commits
+    update_file_without_commit(
+        str(origin_auth_repo.path / "targets/target1"), "invalid_file.txt"
```
renatav, line 625: Same comment, make sure you are modifying the client's repo

taf/tests/test_updater/test_update/test_update_invalid.py
ronantakizawa: I get this error when running update_invalid_repos_and_check_if_repos_exist


E       AssertionError: Regex pattern did not match.
E        Regex: 'Update of (\\w+)\\/(\\w+) failed due to error: Failure to validate (\\w+)\\/(\\w+) commit ([0-9a-f]{40}) committed on (\\d{4}-\\d{2}-\\d{2}): data repository ([\\w\\/-]+) was supposed to be at commit ([0-9a-z]{40}) but repo was at ([0-9a-f]{40})'
E        Input: 'Update of repository failed due to error: Repository test_dirty_index_auth_repo_update_file0/auth has uncommitted changes. Commit and push or revert the changes and run the command again.'

taf/tests/test_updater/conftest.py
ronantakizawa: That was from a previous implementation I don't need it now

renatav (approved):
This looks good now! Great work!

Issue #396, opened 2024-03-13
Configure verbosity to updater output with -v and -vv


n-dusan:
Not all information shown in output for updater is important. Select only the most important messages for updater and show these, while the other suppressed messages are only shown with -v, or -vv flags

renatav:
I've been trying to use different log levels to separate more and less important information. We should be just able to set the debug level.  What that is currently being shown is not important? I thought it made sense to log every step, so that running the updater looks like an installation, where you always know what is happening. So, to me it made sense to print something like "Checking if all existing target repositories are clean..". We could only leave things like "Cloned repository", "Validated repository", "Merged" and error message if it occurs. In that case, we should go over log levels and update where necessary. That being said, there are a lot more messages that are currently hidden, so maybe we can leverage some to a level above info and show the current trace if `-v` flag is provided. Thoughts?

PR #472, opened 2024-07-10
feat: parallelize update clone of dependencies

### LINKED ISSUE #452: Parallelize update/clone of dependencies
There is no reason why we have to update dependencies sequentially. Parallelizing this could speed up the update quite significantly.

jc12690:

Added functionality for parallel execution of child repo updates to improve performance. Utilizes ThreadPoolExecutor for the management of a pool of worker threads for concurrent updates. Fix #452.


taf/updater/updater.py
```diff
@@ -541,9 +545,21 @@ def _update_named_repository(
                         )
                         if error:
                             raise error
+                        return error
```
renatav, line 548: This can only return `None` since it will only be returned if the error is not raised

taf/updater/updater.py
```diff
@@ -541,9 +545,21 @@ def _update_named_repository(
...
                     except Exception as e:
                         errors.append(str(e))
```
renatav, line 550: Since we can no longer append the error to the errors list here, we should return here

taf/updater/updater.py
```diff
@@ -541,9 +545,21 @@ def _update_named_repository(
...
 
+                    # JMC: utilize ThreadPoolExecutor to run the update process in multiple threads
+                    with ThreadPoolExecutor() as executor:
```
renatav, line 553: This is inside of `update_child_repo` and will never be called. I recommend testing this like this:
- create a new directory somewhere, say `test`
- inside that directory, run ` taf repo clone https://github.com/openlawlibrary/open-law --path openlawlibrary/open-law`
- delete test and run clone again to test again

Move this outside of this function, and it will start cloning these dependencies. You'll notice the difference between the outputs

taf/updater/updater.py
```diff
@@ -515,9 +516,12 @@ def _update_named_repository(
...
-                    child_auth_repo
-                ) in child_auth_repos:  # want to parallelize this; separate PR
+                # JMC: Add parallelism for parallel execution of child repo updates to improve performance;
+                # ThreadPoolExecutor manages a pool of worker threads for concurrent updates
+                def update_child_repo(child_auth_repo):
+                    # for (
+                    #        child_auth_repo
+                    # ) in child_auth_repos:  # want to parallelize this; separate PR
                     try:
                         _, error = _update_named_repository(
```
renatav, line 526: Set `out_of_band_authentication` to `out_of_band_authentication=auth_repo.out_of_band_authentication,` here. We need to read this information from the parent repo. One of the first steps will fail without this change

taf/updater/updater.py
```diff
@@ -515,9 +516,12 @@ def _update_named_repository(
                     auth_repo, commits
                 ).values()
 
-                for (
-                    child_auth_repo
-                ) in child_auth_repos:  # want to parallelize this; separate PR
+                # JMC: Add parallelism for parallel execution of child repo updates to improve performance;
+                # ThreadPoolExecutor manages a pool of worker threads for concurrent updates
+                def update_child_repo(child_auth_repo):
+                    # for (
```
renatav, line 522: Please remove all comments when you no longer need them. Only leave those which actually enhance the understanding of the code which follows

renatav (commented):
There are a couple of issues that we will need to resolve. I made the changes I mentioned here and there was still an issue with cloning target repositories. However, before we tackle that, I think that we should implement https://github.com/openlawlibrary/taf/issues/396
The output is now quite useless since the order is random. Ran the command I mentioned and you'll see. I used `log_decorator` to try to make the code cleaner, but I think that we'll need more control here. As the first step, could you replace all `@log_on_start` and `@log_on_end` in `updater_pipeline` with regular logging statements, using `taf_logger`? Then add those flags mentioned in issue #396, just like you added `no-targets` and other flags. Push that when you are done, I'll we'll take it from there

renatav:
Closin in favor of #474

renatav (approved):

Issue #454, opened 2024-06-21
Add support for testing repositories with dependencies to the updater testing framework 


renatav:
These tests cannot be created as easily as other updater tests.

PR #479, opened 2024-07-17
feat: Testing repositories with dependencies

ronantakizawa:

Create-dependencies. Resolves issue #454

### DIFF
--- a/taf/tests/test_updater/test_update_library/test_clone_library.py
+++ b/taf/tests/test_updater/test_update_library/test_clone_library.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+import pytest
 from taf.updater.types.update import UpdateType
 from taf.tests.test_updater.update_utils import (
     check_if_commits_match,
@@ -7,6 +8,23 @@ from taf.tests.test_updater.update_utils import (
 )
 
 
+@pytest.mark.parametrize(
+    "library_with_dependencies",
+    [
+        {
+            "targets_config": [{"name": "target1"}, {"name": "target2"}],
+            "dependencies_config": [
+                {"name": "namespace1/dependency1"},
+                {"name": "namespace2/dependency2"},
+                {"name": "namespace3/dependency3"},
+                {"name": "namespace4/dependency4"},
+                {"name": "namespace5/dependency5"},
+                {"name": "namespace6/dependency6"},
+            ],
+        },
+    ],
+    indirect=True,
+)
 def test_update_repository_with_dependencies(
     library_with_dependencies,
     origin_dir,
--- a/taf/repositoriesdb.py
+++ b/taf/repositoriesdb.py
@@ -108,7 +108,7 @@ def load_dependencies(
         dependencies = dependencies["dependencies"]
         if dependencies is None:
             continue
-
+        print("bruh")
         for name, repo_data in dependencies.items():
             try:
                 if mirrors:
--- a/taf/tests/test_updater/test_update_library/conftest.py
+++ b/taf/tests/test_updater/test_update_library/conftest.py
@@ -1,34 +1,24 @@
 import re
-from taf.repositoriesdb import (
-    DEPENDENCIES_JSON_NAME,
-)
+import pytest
+from taf.repositoriesdb import DEPENDENCIES_JSON_NAME
 from taf.tests.test_updater.update_utils import load_target_repositories
 from taf.tests.test_updater.conftest import (
     RepositoryConfig,
     setup_repository_all_files_initially,
     sign_target_files,
 )
-from taf.tests.conftest import (
-    KEYSTORE_PATH,
-    TEST_INIT_DATA_PATH,
-)
+from taf.tests.conftest import KEYSTORE_PATH, TEST_INIT_DATA_PATH
 from tuf.ngclient._internal import trusted_metadata_set
-from pytest import fixture
 from tuf.repository_tool import TARGETS_DIRECTORY_NAME
-
-
-original_tuf_trusted_metadata_set = trusted_metadata_set.TrustedMetadataSet
 
 NAMESPACE1 = "namespace1"
 NAMESPACE2 = "namespace2"
 TARGET1_NAME = "target1"
 TARGET2_NAME = "target2"
-TARGET3_NAME = "target3"
-AUTH_NAME = "auth"
 ROOT_REPO_NAMESPACE = "root"
+AUTH_NAME = "auth"
 
-
-# test_config.py
+original_tuf_trusted_metadata_set = trusted_metadata_set.TrustedMetadataSet
 
 
 def create_and_write_json(template_path, substitutions, output_path):
@@ -38,7 +28,7 @@ def create_and_write_json(template_path, substitutions, output_path):
     output_path.write_text(template)
 
 
-@fixture
+@pytest.fixture
 def library_with_dependencies(origin_dir):
     library = {}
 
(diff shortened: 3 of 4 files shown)


taf/tests/test_updater/test_update_library/test_clone_library.py
```diff
@@ -7,6 +8,23 @@
 )
 
 
+@pytest.mark.parametrize(
+    "library_with_dependencies",
+    [
+        {
+            "targets_config": [{"name": "target1"}, {"name": "target2"}],
+            "dependencies_config": [
```
renatav, line 16: I think that this is a good start. In `conftest`, instead of iterating over an array of two namespaces, we want to iterate over these dependencies. We want to also be able to specify a `targets_config` inside every `dependencies_config` entry. So, it should also be possible to define which targets each referenced auth repository (dependency) has

renatav (commented):
Haven't tried running yet, but this is looking good! Let's create a couple of tests before merging:
- Valid update of a repository with dependencies
- Update of a repository with dependencies when one referenced authentication repository (dependency) is invalid
- Update of a repository with dependencies when a target repository of a referenced authentication repository is invalid
Check if all repositories got updated, other than the invalid one

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +296,31 @@ def _update_expect_error():
                 assert client_repository.path.exists()
             else:
                 assert not client_repository.path.exists()
+
+
+def invalidate_auth_repo(auth_repo):
```
renatav, line 301: A convention that I try to follow is to put all setup helper function in `conftest`. I think that you should be able to use the setup manager and the same code used to create invalid repositories in `test_clone_invalid`.  When creating the setup manager `setup_manager = SetupManager(auth_repo)` pass in on of the referenced repositories.

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +296,31 @@ def _update_expect_error():
...
+
+    # Corrupt the metadata file
+    root_json_path = auth_repo.path / "metadata" / "root.json"
+    if root_json_path.exists():
+        with open(root_json_path, "w") as f:
+            f.write("corrupted content")
+        print(f"Corrupted root.json at {root_json_path}")
+
+
+def invalidate_target_repo(library_with_dependencies, namespace, target_name):
```
renatav, line 312: Same

renatav (commented):
I really like this. Left a couple of comments, mostly related to some possible code quality improvements. Let's also add update tests. See `test_update_valid` and `set_update_invalid`. The idea here is to run clone, then to run update and check if everything looks good. `clone_full_library` should be renamed to `update_full_library`. It should have operation (clone or update) as its input parameter. See `update_and_check_commit_shas`. 
Great work so far! I'm excited to see these new tests

taf/tests/test_updater/update_utils.py
renatav, line 301: Let's move this to conftest, I'd like all functions which update the repositories to be in the same place.

renatav (commented):
A bit more work left to properly test update. As soon as that is resolved, let's merge this

taf/tests/test_updater/test_update_library/test_clone_library_invalid.py
```diff
@@ -0,0 +1,220 @@
...
+    with pytest.raises(UpdateFailedError, match=INVALID_TIMESTAMP_PATTERN):
+        update_full_library(
+            library_with_dependencies,
+            origin_dir,
+            client_dir,
+            operation=OperationType.CLONE,
+            expected_repo_type=UpdateType.EITHER,
+            excluded_target_globs=None,
+        )
+        update_full_library(
```
renatav, line 66: These `clone` tests should only contain `update_full_library` calls with `operation=OperationType.CLONE`. We want to separate these `clone` and update tests. Let's create two separate modules, `test_update_libray_valid` and `tesrt_update_library_invalid`. There, we could first call `update_full_library` with `operation` set to clone, or create a new function which would just clone without verifying anything (if `test_clone_library_valid` tests pass given the same setup, there is no need to verify everything again). After cloning, we want to make some make some updates (so push new commits to the repositories, valid or invalid, depending on what we are test. These can be the same updates which we are making here, but they need to be executed after the initial clone. So, here we create the repositories, make some changes to them, then clone them using the updater. In the update tests, we want to create the repositories, clone them by running the updater, make some changes and then run `update_full_library` with `operation` set to update

taf/tests/test_updater/test_update_library/test_clone_library_invalid.py
```diff
@@ -0,0 +1,220 @@
...
+            library_with_dependencies,
+            origin_dir,
+            client_dir,
+            operation=OperationType.UPDATE,
+            expected_repo_type=UpdateType.EITHER,
+            excluded_target_globs=None,
+        )
+
+    # Verify that the valid repositories were updated successfully
+    for auth_repo_name, repo_info in library_with_dependencies.items():
```
renatav, line 204: I'd move this to a separate function (`update_utils` could be a place for it). We could check if all valid repositories were updated successfully in every invalid update test. What we should do here is save the head commit shas before the update, run update where we expect an update of one or more dependencies to fail, compare the commits. We should update at least two dependencies - push valid commits to one of them, push invalid commits to another one. After updating the whole library using the update. Then check if head commits shas of repositories to which we pushed valid commits got updated after the update (update the origin repo, the updater should update the client's repositories). This can be a separate issue and not part of this PR. Please create an issue if you'd rather move on to another task.

taf/tests/test_updater/conftest.py
```diff
@@ -665,3 +667,19 @@ def set_head_commit(auth_repo: AuthenticationRepository):
...
+    """
+    Invalidate the given target repository by introducing errors in its structure.
+    """
+    target_repo = next(
+        repo
+        for repo in library_with_dependencies[namespace]["target_repos"]
+        if repo.name == target_name
+    )
+
+    # Remove the .git directory to invalidate the repository
```
renatav, line 682: We could test this, although this doesn't necessarily have to lead to an error (I believe that is the case right now, but that's a big, more so than a security feature). What we really want to be able to do here are 2 things:
- push unauthenticated commits to target repositories. To do so, create and push a commit without signing it (do not update the auth repo), then push again and sign. I also saw that there is an update tests which works in such a way. Let's just add a similar invalid clone test
- create and push an invalid commit to one of the referenced authentication repositories. And I saw that you added this year, awesome!

We can also keep this, we should be able to handle such a case without errors and should address in the future

taf/tests/test_updater/test_update_library/test_clone_library_invalid.py
```diff
@@ -0,0 +1,114 @@
...
+    update_role_metadata_invalid_signature,
+    invalidate_target_repo,
+)
+from taf.updater.types.update import OperationType, UpdateType
+from taf.tests.test_updater.update_utils import (
+    update_full_library,
+)
+
+
+@pytest.mark.parametrize(
```
renatav, line 17: Let's also add a test for the case when the root repo is invalid

taf/tests/test_updater/test_update_library/test_update_library_invalid.py
```diff
@@ -0,0 +1,189 @@
...
+    )
+    # Invalidate one of the authentication repositories in dependencies
+    dependency_auth_repo = library_with_dependencies["namespace1/auth"]["auth_repo"]
+    setup_manager = SetupManager(dependency_auth_repo)
+    setup_manager.add_task(
+        update_role_metadata_invalid_signature, kwargs={"role": "timestamp"}
+    )
+    setup_manager.execute_tasks()
+
+    with pytest.raises(UpdateFailedError, match=INVALID_TIMESTAMP_PATTERN):
```
renatav, line 63: This is exactly what I had in mind!

taf/tests/test_updater/update_utils.py
```diff
@@ -250,9 +284,54 @@ def _update_expect_error():
...
-        # the client repositories should not exits
+        # the client repositories should not exist
         for client_repository in client_repos.values():
             if client_repository.path in repositories_which_existed_paths:
                 assert client_repository.path.exists()
             else:
                 assert not client_repository.path.exists()
+
+
+def validate_updated_repositories(
```
renatav, line 295: You could call this instead of update_full_library in places where we want to invoke the update process and perform these checks

renatav (commented):
This is pretty much done now, just a couple of small updates

taf/tests/test_updater/update_utils.py
```diff
@@ -250,9 +284,54 @@ def _update_expect_error():
...
+    library_with_dependencies,
+    origin_dir,
+    client_dir,
+    valid_target_name,
```
renatav, line 299: These could be arrays of repositories that we expect to get updated successfully, and of repositories which should not get updated. We might also want to check the state of referenced auth repos

renatav (commented):
One minor point I wanted to bring up is that you shouldn't really add any new functionalities to these util functions which call the code you want to test. E.g. raising an error. If we expect the error to be raised, it should be raised by the function we are testing. Which in this case, I am not sure what currently happens if there is a .git dir missing or what we want to happen in such a case. I was going to suggest creating an issue for that and merging this now, but the build failed due to a flake error.

renatav (approved):

Issue #449, opened 2024-06-13
Analyze git error messages


renatav:
- We have very thorough logging, but some `GitError` messages lack detail and we should rethink certain log levels. The problem is that we want to hide a lot of them when running the updater. Related to this issue #396 
- Show full repo paths when showing the error and/or clearly state that the repository does not exist
- Do not ignore this error https://github.com/openlawlibrary/taf/blob/c486a4a2dceab438cd1eb64445a7cea58c470b2f/taf/git.py#L111

PR #485, opened 2024-07-23
fix: checking repo exists and fixing error messages 

rheasenan:
## Checking git repos existence/changing error messages accordingly

Adding an if not statement to check is a git repository and, if true, raising a message that it is not a git repository in git.py. 

Also, deleting old GitError messages that check if the repo is none and has an undescriptive and ill-defined message (about 25 of them).

See #449


taf/git.py
```diff
@@ -28,6 +28,9 @@
 
 EMPTY_TREE = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 
+class PygitError(GitError):
```
renatav, line 31: This should be in `exceptions.py`, you'll see that that is where all of the custom errors are

taf/git.py
```diff
@@ -101,9 +104,14 @@ def __init__(
         self.default_branch = default_branch
 
     _pygit = None
-
+
```
renatav, line 107: Please run the black formatter, or the CI will fail. See https://www.geeksforgeeks.org/python-code-formatting-using-black/

taf/git.py
```diff
@@ -101,9 +104,14 @@ def __init__(
         self.default_branch = default_branch
 
     _pygit = None
-
+    
     @property
     def pygit(self):
+        if not self.is_git_repository:
```
renatav, line 110: This is exactly what I had in mind

taf/git.py
```diff
@@ -101,9 +104,14 @@ def __init__(
...
+            raise GitError(
+                self,
+                message=f"The path '{self.path}' is not a Git repository.",
+            )
```
renatav, line 114: If this error is not raised, but we are still not able to instantiate the pygit repo, raise `PygitError`. There is a `try - expect` here.
``` try:
       self._pygit = PyGitRepository(self)
    except Exception:
        pass
```
raise the error inside that `except Exception` - instead of "swallowing" it.

taf/git.py
```diff
@@ -180,6 +188,8 @@ def log_prefix(self) -> str:
     def pygit_repo(self) -> Optional[pygit2.Repository]:
         try:
             return self.pygit.repo
+        except PygitError as e:
```
renatav, line 191: `self.pygit` will raise an error if `pygit` cannot be instantiated for whatever reason. I don't think that we want to catch it here. Move the logging to `self.pygit` (so log and then raise an error). I'd actually remove this `try-except` - we want the errors raised by `self.pygit` to be propagated.
There should be a test that I sent you over slack, copy-paste it to `test_git` and make sure it's passing

taf/git.py
```diff
@@ -289,11 +299,7 @@ def all_commits_on_branch(
         If branch is None, all commits on the currently checked out branch will be returned
         """
         repo = self.pygit_repo
-        if repo is None:
```
renatav, line 292: This is correct, we don't need these

taf/git.py
```diff
@@ -104,11 +105,18 @@ def __init__(
...
+                message=f"The path '{self.path}' is not a Git repository.",
+            )
+        if self._pygit is None: 
             try:
                 self._pygit = PyGitRepository(self)
-            except Exception:
-                pass
+            except Exception as e:
+                raise PygitError(
+                    "Failed to instantiate PyGitRepository"
```
renatav, line 118: we should include the original error message  - `f" "Failed to instantiate PyGitRepository due to error: {e}"` should work

renatav (approved):

Issue #484, opened 2024-07-23
Revert to last valid state if invalid commits found


renatav:
Currently, if target repositories are out of sync with the authentication repositories, we start the validation from the beginning, and not the last validated commit. If the validation fails, repositories will be updated up to the last valid commit. That should automatically remove all invalid commits from the local repositories.

Create the following tests (some might already exist, we should check):

  1. Client's authentication repository and target repositories are not in sync. Authentication repository contains some commits following the last validated commit. Target repositories are in sync with the last validated commit - simulate the case when someone used `git pull` to update the authentication repository. In fact, that is probably how we want to set this up. Run the updater to clone all repositories (an auth repo and its targets, we should not make this more complicated by testing the update of a repo with dependencies). Make valid changes to both the auth repo and the target repos (only origin repos). Sync client's auth repo with remote without running the updater. Run the updater. Expect and check that all repositories were successfully updated to the last remote commit
2. Client's authentication repository is on last validated commits, but target repositories contain commits following the last validated ones. Simulate the case when someone used `git pull` to update the one or more target repositories.  Run the updater to clone all repositories. Make valid changes to both the auth repo and the target repos (only origin repos). Sync one or more target repos with remote without running the updater. Run the updater. Expect and check that all repositories were successfully updated to the last remote commit.
3. Clients auth repo has been manually updated, just like in 1. The update should result in a partial update. Clients auth repo should be reverted to the last valid commit, last validated commits set to that value, target repositories updated up to commits listed in auth repo's last validated commits
4. Same as 2, but target repositories contain invalid commits following commits which were manually pulled using `git pull`. Expect all repositories to be updated up to the last validated commits and invalid commits to be removed
 5. Client's auth repo is not on last validated commit, but target commits are all over the place. Some are after last validated, some are before. At the moment, i believe that the validation will start from the beginning in such a case and will be successful if there are no invalid commits.

There is some room to improve the performance here:
- if top commits of target repositories are not in sync with the auth repo, but the last validated commits are in git histories of these repositories, start the validation from last validated commit, not from the beginning. 
- if run with `--force`, remove all additional commits from target repositories. That is, revert to last validated commits. This decreases the likelihood that something will go wrong during the update phase

PR #488, opened 2024-07-26
feat: added out of sync tests

ronantakizawa:

Tests repositories that are out of sync and runs manual updates. See #484


taf/tests/test_updater/conftest.py
```diff
@@ -665,3 +668,16 @@ def set_head_commit(auth_repo: AuthenticationRepository):
...
+def sync_auth_repo(auth_repo):
+    # Simulate syncing auth repo with remote
+    auth_repo.pull()
+
+
+def sync_target_repos_with_remote(origin_auth_repo: Path, client_dir: Path):
+    target_repos = list(load_target_repositories(origin_auth_repo).values())
+    for target_repo in target_repos:
+        client_repo = GitRepository(client_dir, target_repo.name)
+        client_repo.fetch()
```
renatav, line 682: pull is a combination of fetch and pull, so you don't have to call fetch first

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,165 @@
...
+            "targets_config": [
+                {"name": "target1", "allow_unauthenticated_commits": True},
+                {"name": "target2", "allow_unauthenticated_commits": True},
+            ],
+        },
+    ],
+    indirect=True,
+)
+def test_auth_repo_not_in_sync(origin_auth_repo, client_dir):
+    #Test 1
```
renatav, line 33: Let's add soe docstrings describing every test case (feel free to copy them from the issue), it's difficult to figure out what we want to test

renatav (commented):
This is pretty much what I had in mind, great work.

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,214 @@
...
+            ],
+        },
+    ],
+    indirect=True,
+)
+def test_auth_repo_not_in_sync_partial(origin_auth_repo, client_dir):
+    """
+    Clients auth repo has been manually updated, just like in 1.
+    The update should result in a partial update.
+    Clients auth repo should be reverted to the last valid commit,
```
renatav, line 119: Sorry, I not explain this too well. What we want to check here is if th updater reverts the local git history to the last commit that was validated. So, add at least one invalid commit and pull manually. The updater should basically remove they invalid commit that was pulled

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_unauthenticated_commits)
+    setup_manager.add_task(update_expiration_dates)
+    setup_manager.add_task(
+        sync_target_repos_with_remote,
+        kwargs={"origin_auth_repo": origin_auth_repo, "client_dir": client_dir},
+    )
+    setup_manager.add_task(
+        revert_last_validated_commit, kwargs={"client_dir": client_dir}
```
renatav, line 167: Same. Here we want to check if the updater removes invalid commits from target repositories that were pulled manually

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,214 @@
...
+                {"name": "target1", "allow_unauthenticated_commits": True},
+                {"name": "target2", "allow_unauthenticated_commits": True},
+            ],
+        },
+    ],
+    indirect=True,
+)
+def test_mixed_target_repo_states(origin_auth_repo, client_dir):
+    """
+    Client's auth repo is not on last validated commit, but target commits are all over the place.
```
renatav, line 188: I probably wanted to say here that auth repo is on last validated commit, but we could test both cases

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,214 @@
...
+    remove_commits(str(client_target_repo_path))
+    clone_repositories(
+        origin_auth_repo,
+        client_dir,
+    )
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.add_task(update_expiration_dates)
+    setup_manager.add_task(
+        remove_commits, kwargs={"repo_path": client_target_repo_path, "num_commits": 1}
```
renatav, line 208: We want the target repos to be completely inconsistent. E.g target 1 is behind last validated commit, target 2 was updated manually

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+    clone_repositories(origin_auth_repo, client_dir)
+    original_commit = origin_auth_repo.head_commit_sha()
+
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(update_expiration_dates)
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.execute_tasks()
+
+    new_origin_commit = origin_auth_repo.head_commit_sha()
+    assert original_commit != new_origin_commit
```
renatav, line 47: We could keep this check, but the most important thing is to verify that the client's repositories are in the correct state. We should add such verification to all of these tests

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.add_task(update_expiration_dates)
+    setup_manager.execute_tasks()
+
+    new_origin_commit = origin_auth_repo.head_commit_sha()
+    assert original_commit != new_origin_commit
+
+    setup_manager.add_task(
+        sync_target_repos_with_remote,
+        kwargs={"origin_auth_repo": origin_auth_repo, "client_dir": client_dir},
```
renatav, line 89: Here we'd like to check that the top commits of target repositories are the same as top commits of origin repos, but not the same as the commit specified in the client's auth repo

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+                {"name": "target2", "allow_unauthenticated_commits": True},
+            ],
+        },
+    ],
+    indirect=True,
+)
+def test_auth_repo_not_in_sync_partial(origin_auth_repo, client_dir):
+    """
+    Clients auth repo has been manually updated, just like in 1.
+    The update should result in a partial update.
```
renatav, line 117: It's important to note that there were invalid commits among the ones that the client pulled

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+        origin_auth_repo,
+        client_dir,
+    )
+
+    # Load the target repositories
+    target_repos = list(load_target_repositories(origin_auth_repo).values())
+    original_commits = {repo.name: repo.head_commit_sha() for repo in target_repos}
+
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_unauthenticated_commits)
```
renatav, line 132: We should add invalid commits here. Take a look at invalid clone and update tests

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+    clone_repositories(
+        origin_auth_repo,
+        client_dir,
+    )
+    target_repos = list(load_target_repositories(origin_auth_repo).values())
+    original_commits = {repo.name: repo.head_commit_sha() for repo in target_repos}
+
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_unauthenticated_commits)
+    setup_manager.execute_tasks()
```
renatav, line 173: Add valid authenticates target commits following unauthenticated ones. That will result in an invalid update. Should we combine the two to make it easier to set up this invalid state?

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+
+    update_and_check_commit_shas(OperationType.UPDATE, origin_auth_repo, client_dir)
+
+    # revert target1, add to target2
+    setup_manager.add_task(
+        remove_commits, kwargs={"repo_path": reverted_repo.path, "num_commits": 1}
+    )
+    setup_manager.add_task(
+        update_target_files,
```
renatav, line 228: I think that this just updates the auth repo's target files and not the actual target repo. We could just manually pull again (but only one target repo)

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,294 @@
...
+        },
+    )
+    setup_manager.execute_tasks()
+
+    # Check that the manually updated repo has a new commit
+    manually_updated_commit = manually_updated_repo.head_commit_sha()
+    assert original_commits[manually_updated_repo.name] != manually_updated_commit
+
+    # Check that the untouched repo has the same commit
+    untouched_commit = untouched_repo.head_commit_sha()
```
renatav, line 289: These checks are great, that's exactly what o had in mind!

renatav (commented):
It's great that none of these failed. A couple of relatively minor things to address.

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,326 @@
...
+    assert original_commits[updated_repo.name] != updated_commit
+
+    # Sync the authentication repo
+    setup_manager.add_task(
+        sync_target_repos_with_remote,
+        kwargs={"origin_auth_repo": origin_auth_repo, "client_dir": client_dir},
+    )
+    setup_manager.execute_tasks()
+
+    verify_client_repos_state(client_dir, origin_auth_repo)
```
renatav, line 257: Should call the updater here. Given this particular setup, I believe that the updater will false without the --force flag?

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,326 @@
...
+    untouched_commit = untouched_repo.head_commit_sha()
+    assert original_commits[untouched_repo.name] == untouched_commit
+
+    # Revert the manually updated repository to its original commit
+    setup_manager.add_task(
+        sync_target_repos_with_remote,
+        kwargs={"origin_auth_repo": origin_auth_repo, "client_dir": client_dir},
+    )
+    setup_manager.execute_tasks()
+    # update_and_check_commit_shas(OperationType.UPDATE, origin_auth_repo, client_dir)
```
renatav, line 324: Similarly, I believe that this will fail without the `--force` flag. What I was thinking is that we should manually update some of the client's target repositories and remove one or more commits from some other ones. There is no need to add additional unsigned commits. In that case, the updater should start the validation from the beginning. I can't think of a quick and elegant was of testing that if the validation started from the beginning or not. The simplest way could be to enable file logging and check the log, but I don't think that we need to do that as a part of this PR

taf/tests/test_updater/test_update_library/test_validation_and_sync.py
```diff
@@ -0,0 +1,306 @@
...
+    original_commit = origin_auth_repo.head_commit_sha()
+
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(update_expiration_dates)
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.execute_tasks()
+
+    new_origin_commit = origin_auth_repo.head_commit_sha()
+    assert original_commit != new_origin_commit
+    origin_auth_repo.pull()
```
renatav, line 48: We need to update the local (client's) repositories. `origin_auth_repo` is updated by running the `SetupManager`, there is no need to pull the changes again. We need to simulate various cases when the users manually update their local repositories (by running git pull). Instead of running `origin_auth_repo.pull()`, we need to to update repositories inside `client_dir`. Their paths can be calculated by combining that path with the names of the repositories. Please update all of these tests. Sorry for not pointing that out sooner

renatav (commented):
We'll see if tests will start failing after these updates. Also, let's move these tests to `test_update`. We are not testing the update of a repository with dependencies.

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +257,112 @@ def _update_expect_error():
...
+        origin_repo = GitRepository(origin_auth_repo.path.parent.parent, repo_name)
+        client_commit = client_repo.head_commit_sha()
+        origin_commit = origin_repo.head_commit_sha()
+
+        # Assert that the top commits of target repositories are the same as the top commits of origin repositories
+        assert (
+            client_commit == origin_commit
+        ), f"Target repo {repo_name} should have the same top commit as the origin repo"
+
+        # Assert that the top commits of target repositories are not the same as the commit specified in the client's auth repo
```
renatav, line 365: To check if top commits or target repositories are the same as the commits specified in the auth repo, read the expected commit shas from target files: `
```
  target = client_auth_repo.get_target(repo_name)
  commit_sha = target["commit"]
  assert commit_sha ==  client_commit
```

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +257,112 @@ def _update_expect_error():
...
+                client_commit != origin_commit
+            ), f"Target repo {repo_name} should not have the same commit as specified in the client's auth repo"
+
+
+def verify_partial_update(
+    client_dir: Path, origin_auth_repo: AuthenticationRepository, original_commits: dict
+):
+    """
+    Verify that the client's repositories are in the correct state following a partial update.
+    This means that the top commits of target repositories are the same as the top commits of origin repositories,
```
renatav, line 344: The opposite should be true. Top commits of the client's local target repositories, in case of a partial update, should be different from the top commits of the origin repos (partial update is an unsuccessful update, which failed because the origin repos contained invalid commits). If a user manually pulled invalid commits, these tests need to verify that the the updater is able to remove them. Repositories should be at the most recent valid commit

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +257,112 @@ def _update_expect_error():
...
+            target_file_path = client_repo.path / "test1.txt"
+
+            # Ensure the target file exists
+            if not target_file_path.is_file():
+                raise FileNotFoundError(
+                    f"Target file {target_file_path} does not exist"
+                )
+
+            # Extract commit SHA from the target file in the client repo
+            target_commit_sha = client_repo._git(
```
renatav, line 326: You can use  `client_auth_repo.get_target(repo_name)` (see the last comment).

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +257,112 @@ def _update_expect_error():
...
+        for repo_name, client_repo in client_target_repos.items():
+            origin_repo = GitRepository(auth_repo.path.parent.parent, repo_name)
+            client_commit = client_repo.head_commit_sha()
+
+            # Get the target file path in the client repo
+            target_file_path = client_repo.path / "test1.txt"
+
+            # Ensure the target file exists
+            if not target_file_path.is_file():
+                raise FileNotFoundError(
```
renatav, line 321: You usually don't raise errors in tests. If you want the test to fail if the file does not exist, use assert:
`assert target_file_path.is_file()`

taf/tests/test_updater/update_utils.py
```diff
@@ -256,3 +257,112 @@ def _update_expect_error():
...
+    client_repos = {
+        repo.name: GitRepository(client_dir, repo.name)
+        for repo in load_target_repositories(origin_auth_repo).values()
+    }
+    # Add the origin authentication repository itself to the client repositories
+    client_repos[origin_auth_repo.name] = GitRepository(
+        client_dir, origin_auth_repo.name
+    )
+
+    try:
```
renatav, line 281: Catching assertion errors is not considered to be good practice. If you need to reuse the check, create a helped function without an assert.

taf/tests/test_updater/conftest.py
```diff
@@ -681,3 +681,11 @@ def set_head_commit(auth_repo: AuthenticationRepository):
         auth_repo.set_last_validated_commit(last_valid_commit)
     else:
         raise ValueError("Failed to retrieve the last valid commit SHA.")
+
+
+def pull_client_repos(auth_repo: AuthenticationRepository, client_dir: Path):
```
renatav, line 686: Something like this, yes. We, however, need to have the flexibility to run pull for just one target repo, all target repos and client's auth repo. Create client's auth repo appending origin auth repo's name to client_dir.  Exactly what we want to do depends on the test. This can be split into multiple setup functions, whatever you think it's better

renatav (approved):
The remaining issues will be addressed in a separate PR

Issue #486, opened 2024-07-23
Update when auth repo's top commit is behind last validated commit


renatav:
At the moment, I believe that update will fail if the last validated commit is not in the repository's git history. It's a lot trickier when the top commit is behind the last validated commit, as it is not as easy to verify that the last validated commit is actually valid. Something like this can happen when someone runs the updater. The updater will update last validated commit (content of the `last_validated_commit` file) automatically. If the user then removes some commits from the repo, the top commit of their local repository will be behind the last validated commit. We could do the following:
- check if the top commit of user's local repository is a valid commit by checking if it is contained by the remote repo
- check if `last_validated_commit` is a valid commit by checking if it is contained by the remote repo
- check if top commits of target repositories are in sync with the top commit of the auth repo (or if commits defined in auth repo are in their histories)
- if that is all true, start the update from the top commit.
- if commits of target repositories are all over the place, start the update from the first commit
- If `last_validated_commit` is not contained by the remote repo, that could mean that someone force pushed removed commits (we currently do this if the repository somehow ends up in an invalid state, we do not have a rollback option or revocations - we need to implement them)
  - when called without `--force` raise an error
  - when called with `--force`, set last validated commit to top commit of the repo, if it is contained by the remote repo and set commits of target repositories to the corresponding commits if none of the target repos are behind the top auth commit. Start update from that commit
  -  if some of these requirements are not met, raise an error even with force. The user might want to clone the repo again at this point, but let them deal with that

Issue #210, opened 2022-02-15
Handling different states 


n-dusan:
### Necessary for TAF 1.0

- [x] #496

#### Different states in TAF could be:

```
args:
    - `--force` - force all repositories into consistent state at LVC and then run updater; may lose data. provide info on all changes made to repositories in order get to consistent state (use reflog if needed)


LVC = Last Validated Commit

axis:
    - working directory
        - clean  
            - happy path
        - ~dirty~
            - DEFAULT: FAIL
            - --force: git clean -fd; git reset HEAD --hard; proceed with update
    - branch
        - correct
            - happy path
        - incorrect branch or detached HEAD?
            - DEFAULT: stay on the incorrect branch; update correct branch; provide warning that not on the correct branch
            - --force: switch to LVC branch; proceed with update
    - commit
        - at last validated commit
            - happy path
        - validatable (in authentication chain, even if it is in "future" relative to last validated commit; this could happen if you manually pull or reset --hard)
            - DEFAULT: update to LVC; proceed with update
        - not validatable (not in validation chain)
            - DEFAULT: FAIL
            - --force: force branch to LVC; proceed with update
        - trailing unvalidated commits (in law-xml-codified but not law) - we call them unauthenticated in updater
            - DEFAULT: FAIL
            - --force: reset LVC --hard; proceed with update
    - local repositories (all commits are consistent at repository level)
        - doesn't exist, no commit
            - DEFAULT: create repo; ff to LVC
    - local vs remote
        - remote is consistent with local - no trailing unauthenticated commits
            - happy path
        - remote is consistent with local - trailing unauthenticated commits on remote 
            - DEFAULT: update to last validatable commit on remote; then fail for any unauthenticate commits on repos that don't allow; notify of commits on repos that do allow
        - remote is consistent with local - trailing unauthenticated commits on local
            - DEFAULT: update to the last authenticatable commit on remote; FAIL if repo does not allow trailing commits or do nothing if repo allows trailing commits
            - --force: remove trailing commits
        - remote is consistent with local - trailing unauthenticated commits conflict on repo that allows trailing commits
            - DEFAULT: update to the last authenticatable commit on remote; FAIL
            - --force: update to the last authenticatable commit on remote; remove local commits
        - remote is inconsistent with local (i.e. both remote and local are valid but diverge)
            - DEFAULT: update to the last authenticable commit on remote; FAIL
        - remote state is not authenticatable (metadata issues, versions, keys, signing, etc.)
[...]

PR #490, opened 2024-07-30
feat: Update when auth repo's top commit is behind last validated commit

ronantakizawa:

Updates a repo behind last validated commit using `taf repo update --force`. See #486 and #502 


taf/updater/updater_pipeline.py
```diff
@@ -657,7 +657,13 @@ def _validate_last_validated_commit(self, last_validated_commit):
...
                 msg = f"Top commit of repository {self.state.users_auth_repo.name} {users_head_sha} and is not equal to or newer than last successful commit"
                 taf_logger.error(msg)
-                raise UpdateFailedError(msg)
+                if self.force:
+                    taf_logger.info(
+                        "Forcing update: setting last validated commit to the top commit of the local repository."
+                    )
+                    self.state.users_auth_repo.set_last_validated_commit(users_head_sha)
+                else:
+                    raise UpdateFailedError(msg)
```
renatav, line 666: The problem here is that we don't really know if the commit is even valid. I don't think there is any other way to check that without checking if the remote repo contains it.

taf/updater/updater_pipeline.py
```diff
@@ -670,19 +670,65 @@ def validate_out_of_band_and_update_type(self):
...
+        def validate_commit_in_remote(repo, commit_sha):
+            try:
+                repo.fetch()
+                remote_commits = repo.all_commits_on_branch(repo.default_branch)
+                return commit_sha in remote_commits
+            except GitError:
+                return False
+
+        # Check if the top commit of the user's local repository is valid
+        if not validate_commit_in_remote(self.state.users_auth_repo, users_head_sha):
```
renatav, line 683: We do not want to make any changes to the user's local repo before we validate the changes. We should use the temp repos instead (we clone all repositories to temp, validate them, and then update the user's local repositories)

taf/updater/updater_pipeline.py
```diff
@@ -670,19 +670,65 @@ def validate_out_of_band_and_update_type(self):
     def _validate_last_validated_commit(self, last_validated_commit):
         branch = self.state.users_auth_repo.default_branch
         users_head_sha = self.state.users_auth_repo.top_commit_of_branch(branch)
-        if last_validated_commit != users_head_sha:
```
renatav, line 673: All following checks are only needed if these two commits are not the same, so we should probably keep this. Also, check if the last validated commit is in that git repo's history first, as that should be quicker than running fetch

taf/updater/updater_pipeline.py
```diff
@@ -670,19 +670,65 @@ def validate_out_of_band_and_update_type(self):
...
+                else:
+                    taf_logger.error(msg)
+                    raise UpdateFailedError(msg)
+            else:
+                taf_logger.info(
+                    f"Commit {users_head_sha} is a valid successor of the last validated commit {last_validated_commit}."
+                )
+
+        # Handle the case where there is no last validated commit
+        if not last_validated_commit:
```
renatav, line 726: When there is no last validated commit, we should start the validation from the first commit. I believe that is how it works on master

renatav (commented):
We should create a couple of tests before:
- Clone the repositories using the updater. Make sure that there are a couple of commits. Use `reset --hard` to remove one or more commits from the auth repo. Run the updater again.
- Clone the repositories using the updater. Push a couple of commits to the origin repo. Update the last validated commit manually, set it to one of the new commits. This should ensure that there are the last validated commit is valid, but not contained by the user's local repository.
- Manually set the last validated commit to an invalid commit and run the update.

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -639,3 +643,96 @@ def test_update_partial_with_invalid_commits(origin_auth_repo, client_dir):
...
+    # Push new commits to the origin repo
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.execute_tasks()
+
+    # Manually set the last validated commit to one of the new commits
+    client_auth_repo = AuthenticationRepository(client_dir, origin_auth_repo.name)
+    last_commit_sha = client_auth_repo.last_validated_commit
+    client_target_repos = load_target_repositories(client_auth_repo)
+    origin_auth_repo.set_last_validated_commit(last_commit_sha)
```
renatav, line 707: Something like this, but it's the client's repo whose last validated commit should be updated. The updater is used to update the client's repositories and sync them with the origin repos. Think of origin repos hosted by GitHub.

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -639,3 +643,96 @@ def test_update_partial_with_invalid_commits(origin_auth_repo, client_dir):
...
+    ],
+    indirect=True,
+)
+def test_update_with_invalid_last_validated_commit(origin_auth_repo, client_dir):
+    # Step 1: Clone the repositories
+    clone_repositories(origin_auth_repo, client_dir)
+
+    # Step 2: Set the last validated commit to an invalid commit (e.g., non-existent SHA)
+    invalid_commit_sha = "0000000000000000000000000000000000000000"
+    origin_auth_repo.set_last_validated_commit(invalid_commit_sha)
```
renatav, line 735: Same. The update might fail

taf/updater/updater_pipeline.py
```diff
@@ -668,21 +668,59 @@ def validate_out_of_band_and_update_type(self):
...
+            first_commit_sha = validation_repo.get_first_commit_on_branch(branch)
+            last_validated_commit = first_commit_sha
+
+        # Check if the last validated commit is in the temp repo's history first
+        if last_validated_commit not in validation_repo.all_commits_on_branch(branch):
+            # If not in temp history, check if it's in the remote history
+            if not validate_commit_in_remote(validation_repo, last_validated_commit):
+                msg = f"Last validated commit {last_validated_commit} is not in the remote repository."
+                if self.force:
+                    if validate_commit_in_remote(validation_repo, users_head_sha):
```
renatav, line 696: I think that this should all work without the force flag, since we're not modifying the user's (client's) repo. The temp repo should contain all remote commits, since we clone and fetch from remote.

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -730,9 +731,13 @@ def test_update_with_invalid_last_validated_commit(origin_auth_repo, client_dir)
     # Step 1: Clone the repositories
     clone_repositories(origin_auth_repo, client_dir)
 
-    # Step 2: Set the last validated commit to an invalid commit (e.g., non-existent SHA)
-    invalid_commit_sha = "0000000000000000000000000000000000000000"
-    origin_auth_repo.set_last_validated_commit(invalid_commit_sha)
+    # Step 2: Delete the last validated commit file to simulate an invalid commit (non-existent SHA)
```
renatav, line 734: The update should be successful if there is not last validated commit. We expect it to start from th beginning. We should definitely keep this test, but also need to test the case when last validated commit is set to an invalid commit. The updater might fail. If that is the case, just comment out that test and I'll work on that later

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -34,6 +37,7 @@
     indirect=True,
 )
 def test_update_valid_happy_path(origin_auth_repo, client_dir):
+    print("Cloning repositories")
```
renatav, line 40: Let's remove these print statements before merging

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -639,3 +644,125 @@ def test_update_partial_with_invalid_commits(origin_auth_repo, client_dir):
...
+    clone_repositories(origin_auth_repo, client_dir)
+
+    # Push new commits to the origin repo
+    setup_manager = SetupManager(origin_auth_repo)
+    setup_manager.add_task(add_valid_target_commits)
+    setup_manager.execute_tasks()
+
+    # Manually set the last validated commit to one of the new commits
+    client_auth_repo = AuthenticationRepository(client_dir, origin_auth_repo.name)
+    last_commit_sha = client_auth_repo.last_validated_commit
```
renatav, line 706: Set th commit to the top commit of the origin repo. That should ensure that the commit exists and is valid, but does not exist in the client's repo. This is not very likely to happen, but we need to handle such a case too

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -639,3 +644,125 @@ def test_update_partial_with_invalid_commits(origin_auth_repo, client_dir):
...
+    )
+    if last_validated_commit_file.exists():
+        last_validated_commit_file.unlink()  # Remove the file
+
+    # Step 3: Run the updater and expect it to pass by handling the missing last validated commit
+    update_and_check_commit_shas(OperationType.UPDATE, origin_auth_repo, client_dir)
+
+
+"""
+Test Fails
```
renatav, line 747: I've given it some more thought and I think that this test should fail. The user might need to resolve this manually. With the --force flag, we could set the commit to the top commit of the auth repo (if it is valid, check remote). We could move this to test_invalid.

renatav (commented):
Could you also create some documentation based on the issues and how we handled these various cases (not just this one, but all of the issues starting with the addition of the force flag)? That would be very useful

docs/updater/update_process.md
```diff
@@ -158,3 +159,12 @@ if __name__ == '__main__':
     state = do_something(data)
     send_state(state)
 ```
+
+
+# Forced Updates
+
+A forced update overrides the normal update process to address situations where the authentication repository or its target repositories have encountered inconsistencies that prevent a standard update. During a forced update, the updater bypasses certain checks that would normally stop the process, such as the presence of an invalid or missing last validated commit. The forced update ensures that the repository is aligned with the latest commit found on the remote repository. 
+
+By using the --force flag, the updater will reset the state of the repository, discarding any uncommitted changes, and align the repository with the latest commit from the remote repository. This includes resetting the last validated commit to the latest commit found on the remote, and synchronizing the target repositories accordingly. If the target repositories have diverged or contain unaligned commits, the updater will reset them to the validated state, potentially discarding local changes.
```
renatav, line 168: The updater should not set the last validated commit to a commit with might have not validated yet. The reason why we are doing so in a test is to simulate a case when commits were removed from the local repository manually, or when someone manually updated the last validated commit. Force flag can remove certain commits and uncommitted charges

docs/updater/updater.md
```diff
@@ -25,3 +25,7 @@ For the above listed reasons, we cannot simply use TUF's updater like described
...
 
 - Since TUF does not provide a way to gradually update metadata and target files (except for `root.json`), that has to be handled by TAF. So, TAF's updater calls TUF's updater multiple times. Our handler, `GitUpdateHandler` finds all commits that were pushed after the last time the client pulled the changes. The basic idea is to traverse trough these commits and update metadata files and targets for each of them, thus validating all revisions.
+
+- Validation and syncing before running the updater is crucial for the updater to work. Using `git pull`, we can ensure that both the authentication repository and target repositories are aligned with their respective last validated states. The process involves comparing the current state of the repositories on the client's machine with the latest state on the remote server and making necessary adjustments to bring them into sync. Examples of when validation and syncing is necessary when the client's authentication repository is not in sync with the remote repository, when target repositories are out of sync, when there are invalid or unauthorized commits, when repositories are in mixed states, and after manual updates.
```
renatav, line 29: I don't think that we should ever manually run gut pull before running th updater. The issue being addressed in this PR is a lot trickier, so instead of focusing on it, let's document what we did as a part of previous PRs (removal of additional commits, uncommitted changes etc.)

docs/updater/updater.md
```diff
@@ -25,3 +25,7 @@ For the above listed reasons, we cannot simply use TUF's updater like described
...
+
+- Validating the last validated commit in the updater involves ensuring that the commit being used as a reference for updates is still valid and consistent with the current state of the repository. The process starts by identifying the main branch of the repository and obtaining the latest commit on that branch. If no specific commit is provided for validation, the process defaults to using the first commit in the branch's history. Next, the updater checks whether the provided last validated commit is still part of the branch's commit history. If the commit is missing, the updater fetches the latest commits from the remote repository to verify if the commit exists there. If the commit is found but the repository's current state has moved beyond this commit without aligning properly, the updater ensures that the last validated commit is updated to match the latest commit from the remote repository, guaranteeing that the validation process is based on the most recent and accurate state of the repository.
```
renatav, line 31: Same here's we should never set the last validated commit to a commit that has not yet been fetched. If th Last validated commit is already set to a commit that is not contained by the local repo, but is by remote one, we don't have to raise an error.

renatav (commented):
Let's document this issue https://github.com/openlawlibrary/taf/issues/484
I am much more confident that we handled all of those cases correctly. Also, the force flag which removes uncommitted changes and some commits. In addition to updating these existing documents, could you create a new one (say, states.md inside the updater docs directory) and just list what we do to resolve some of these inconsistencies? 
Something inspired by https://github.com/openlawlibrary/taf/issues/210

So that we don't have to keep referring to that issue.

docs/updater/states.md
```diff
@@ -0,0 +1,84 @@
+### Necessary for TAF 1.0
```
renatav, line 1: Let's just remove this heading

renatav (approved):

Issue #244, opened 2022-05-04
Use multiple YubiKeys at the same time


renatav:
Check if we can handle having multiple inserted YubiKeys - e.g. one containing snaphost and one containing timestamp key. Should be able to load both keys.
If there is ever a need to specify which Yubikey is to be used, make sure they can be done in a user-friendly way

ronantakizawa:
Using multiple Yubikeys at the same time is not supported for all devices (macOS). 

A way to work around this is to send certain keys to 1 Yubikey and other to another yubikey

renatav:
Note that #503 was merged into a feature branch

PR #503, opened 2024-08-14
feat: Use multiple yubikeys at same time

ronantakizawa:

Use multiple yubikeys at same time (Key generation, repo generation, target signing). Resolves issue #244 


taf/repository_tool.py
```diff
@@ -146,27 +146,35 @@ def yubikey_signature_provider(name, key_id, key, data):  # pylint: disable=W061
...
-            pin = yk.get_key_pin(serial_num)
+            yubikey_count = yk.check_yubikey_count()
+            if yubikey_count < 2:
+                inserted_key = yk.get_piv_public_key_tuf()
+                if expected_key_id != inserted_key["keyid"]:
+                    return None, None
+                serial = yk.get_serial_num(inserted_key)
+            else:
+                print(f"Final confirmation for {name} key")
+                serial = yk.verify_yubikey_serial()
```
renatav, line 157: I am currently unable to test this since I am not able to insert two yubikeys at the same time, but based on the code, it looks like you have to manually enter each of the yubikey's serial number. This is probably less user-friendly compared to having to remove and insert the devices. A bigger issue, however, is that we want to be able to sign metadata using YubiKeys plugged into our server. So, this should iterate over all inserted YubiKeys and simply try to sign using all of them, instead of asking for any input. So, let's say that you have timestamp and snapshot keys inserted and you need to sign timestamp metadata. We need to automatically detect the key that can sign the timestamp metadata. At the moment, the user will have to enter pins too, but we don't need to address that in this PR.

ronantakizawa:
Currently only works if you have a key-description.json file since we use the public key to check which Yubikey is which.

renatav (approved):
I'll merge this into a feature branch, to move this code off of the fork. Makes sense at first glance, but I am not able to test this since I am not able to plug in two YubiKeys at the same time.

PR #514, opened 2024-08-27
fix: move yubikey_utils from tests folder

sale3:

`yubikey_utils` module should be moved from tests folder since this folder is excluded during wheel creation.
Use random generator suitable for security/cryptographic purposes (bandit).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ except ModuleNotFoundError:
     tests_exist = False  # type: ignore
 if tests_exist:
     kwargs["entry_points"]["pytest11"] = (
-        ["taf_yubikey_utils = taf.tests.yubikey_utils"],
+        ["taf_yubikey_utils = taf.tools.yubikey.yubikey_utils"],
     )
 
 setup(**kwargs)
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -105,7 +105,7 @@
     tests_exist = False  # type: ignore
 if tests_exist:
     kwargs["entry_points"]["pytest11"] = (
-        ["taf_yubikey_utils = taf.tests.yubikey_utils"],
+        ["taf_yubikey_utils = taf.tools.yubikey.yubikey_utils"],
     )
```
n-dusan, line 109: question: do we still need this code? If `config.pluginmanager` on platform is the only place we're using it and if `taf.tools.yubikey.yubikey_utils` works, could we use that instead? Then we can remove `taf_yubikey_utils` entrypoint from `setup.py`

sale3: Checked, works with just importing `yubikey_utils` from the new path. I will create another PR on platform to remove the plugin.
update: I will dismiss this PR since it was checked out from wrong tag. I will create a new one.

PR #542, opened 2024-09-27
Issue 501

yuktakul04:


tests/test_ykman.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+try:
+    from ykman.device import list_all_devices
+    YKMAN_AVAILABLE = True
+except ImportError:
+    YKMAN_AVAILABLE = False
+
+@pytest.mark.skipif(not YKMAN_AVAILABLE, reason="ykman module is not available")
```
renatav, line 8: I think this `skipif` will work, but `test_repository_tool` should be updater and `taf.yubikey` should probably be surrounded by a similar try-except.

yuktakul04: I have edited the code accordingly. Hope this seems reasonable.

renatav (commented):
Instead of adding a new tests, `test_repository_tool` needs to be updated. Tests that require ykman should be skipped. There should be no failing tests if there is ykman is not installed

renatav:
So what we need to do here is update `taf\taf\tests\test_repository_tool\test_repository_tool.py`. If you install `yubikey-manager` dependency, (install taf by running `pip install -e .[yubikey]` from the root of the repository), you should be able to run this test using `pytest`. Then, uninstall yubikey-manager: `pip uninstall yubikey-manager`. You see this error:
```
ImportError while loading conftest 'D:\oll\taf\taf\tests\test_repository_tool\conftest.py'.
conftest.py:8: in <module>
    from taf.tools.yubikey.yubikey_utils import (
..\..\tools\yubikey\__init__.py:3: in <module>
    from taf.api.yubikey import export_yk_certificate, export_yk_public_pem, get_yk_roles, setup_signing_yubikey, setup_test_yubikey
..\..\api\yubikey.py:14: in <module>
    import taf.yubikey as yk
..\..\yubikey.py:17: in <module>
    from ykman.device import list_all_devices
E   ModuleNotFoundError: No module named 'ykman'
```
If you take a closer look at it, there is a `import taf.yubikey as yk` in this module, but also there are imports from `taf.tools.yubikey.yubikey_utils` in that `conftests` (conftest in the same package). The easiest way to fix this would be to create a new package, say `test_repository_tool_w_yubikey` or something like that. Create a `conftest` inside that package and then move everything that requires a yubukey to that `conftest`. E.g. fixtures like 
```
@fixture
def targets_yk(pytestconfig):
    """Targets YubiKey."""
    return TargetYubiKey(KEYSTORE_PATH, pytestconfig.option.signature_scheme)
```
Move everything from `test_repository_tool` to that a module inside this new package. Then skip the whole module if the dependency is not installed.

PR #551, opened 2024-10-16
fix: use mirrors.json urls when cloning dependencies

n-dusan:

Fixes NILL when running `taf repo clone https://github.com/oll-test-repos/narf-nill-law.git` and some of the target repos have ssh instead of https urls.

### DIFF
--- a/taf/updater/updater.py
+++ b/taf/updater/updater.py
@@ -139,8 +139,9 @@ def _reset_to_commits_before_pull(auth_repo, commits_data, targets_data):
 @define
 class UpdateConfig:
     operation: OperationType = field(converter=OperationType)
-    url: str = field(
-        metadata={"docs": "URL of the remote authentication repository"}, default=None
+    remote_url: str = field(
+        metadata={"docs": "Remote URL of the remote authentication repository"},
+        default=None,
     )
     path: Path = field(
         default=None,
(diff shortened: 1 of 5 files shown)


taf/updater/updater.py
```diff
@@ -139,8 +139,9 @@ def _reset_repository(repo, commits_data):
 @define
 class UpdateConfig:
     operation: OperationType = field(converter=OperationType)
-    url: str = field(
-        metadata={"docs": "URL of the remote authentication repository"}, default=None
+    remote_url: str = field(
```
renatav, line 142: Why do we need to separate remote and clone urls? We want to attempt to clone the repo from all remote urls

n-dusan: I believe it's related to `taf repo clone` only taking in a single URL, that is then passed down to all child dependent repositories as well. It appears that we have [this](https://github.com/openlawlibrary/taf/blob/692c469ec7328f5eee9eeef23e25366e28e1b42e/taf/updater/updater.py#L484) and it does not seem very flexible in case of dependent authentication repositories. 

Do you have a particular solution in mind? If not, I'd vote we sort out in a separate issue.

renatav: So `clone` iterates over all urls and attempts to clone from each of them, until it succeeds in cloning the repo. The reason why there's only one auth repo url is that when you run `taf repo clone` you only specify one url. I never really put too much thought into cloning dependencies. I vote for replacing this existing url with a list of urls.

renatav (approved):

PR #559, opened 2024-10-24
Fully support partial update 

### LINKED ISSUE #251: Fully support partial update
In some cases, it would make sense to only validate and update the authentication repository and some target repositories. However, the success of such an update should not result in updating `last_validated_commit` since we cannot know if there exists a problem with an excluded target repository. On the other had, it would be nice to skip validation of the already validated repositories the next time update is run, and only check those which were previously excluded. This additional information can be stored in the config directory and loaded when the next update is started.

renatav:

Fully supporting partial update means adding support for excluding certain target repositories and storing information about what was validated and up to which auth commit. `last_validated_commit` is only updated if clone and update are run without excluding any target repositories. We have no way of knowing if the whole set of repos is valid unless all target repos are validated too. This PR introduces a new file, called `last_validated_data`, where we store last validated commit for each repository. These repositories will only be updated from those commits. The reason why I didn't opt for storing everything in this file is backwards compatibility. This all makes the update process a bit more complicated because different repos are validated from different auth commits.

Closes #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,57 @@ and this project adheres to [Semantic Versioning][semver].
 [559]: https://github.com/openlawlibrary/taf/pull/558
 
 
+## [0.32.4]
+
+### Added
+
+### Changed
+
+- Change log level for `repositoriesdb` messages ([569])
+
+### Fixed
+
+[569]: https://github.com/openlawlibrary/taf/pull/569
+
+## [0.32.3] - 11/22/2024
+
+### Added
+
+### Changed
+
+### Fixed
+
+- Fix `get_last_remote_commit` - add missing value for parameter ([566])
+
+[566]: https://github.com/openlawlibrary/taf/pull/566
+
+## [0.32.2] - 11/20/2024
+
+### Added
+
+### Changed
+
+- Make url optional for `get_last_remote_commit` ([564])
+
+### Fixed
+
+
+[564]: https://github.com/openlawlibrary/taf/pull/564
+
+## [0.32.1] - 11/01/2024
+
+### Added
+
+### Changed
+
+### Fixed
+
+- Fix two git methods where `GitError` was not being instantiated correctly ([562])
+- Fix determination of auth commits to be validated when starting the update from the beginning ([562])
+
+[562]: https://github.com/openlawlibrary/taf/pull/562
+
+>>>>>>> master
 ## [0.32.0] - 10/23/2024
 
 ### Added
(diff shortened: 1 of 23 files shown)


n-dusan (commented):
I've pushed a fix to get lifecycle scripts running and tested it with the narf-nill/law repo. It’s working great!

I do have a thought here, and I’d love your input if you have any thoughts or suggestions. No worries if you think this would be overly complex to modify. We can always go ahead and merge it as-is.

> The reason why I didn't opt for storing everything in this file is backwards compatibility. This all makes the update process a bit more complicated because different repos are validated from different auth commits.

Just wondering: Are we using `last_validated_commit` in our workflows? If not, might it simplify the code base to maintain just one file, like` last_validated_data` or `last_validated_commit`? My understanding is that the current LVC setup mostly applies to our internal team. If so, could we address backward compatibility by asking them to delete the `_law` directory before updating?

renatav:
@n-dusan If I remember correctly, there were a couple of components that were reading from this file. You're probably more familiar with the changes that were made in the last two years, so that might not be the case. I'll let you make this call

n-dusan:
@renatav Thanks for the context! I’m not entirely sure if anything still reads from this file either. Happy to chat over it if you think that’d be helpful. Otherwise, I’ll dig a bit more and keep you posted.

renatav:
@n-dusan Whatever you think would be easier. I just don't want there to be a fire down the line because LVC's structure was updated, leading to stressful debugging sessions. That is the only reason why I decided to introduce a second file.

renatav:
@n-dusan This is ready for another review

n-dusan (commented):
I ran the updater again and got confused as to why we still keep `last_validated_commit` in the file. Figured it's unnecessary and I've pushed a fix to remove it in this commit - [fix: remove last_validated_commit entry](https://github.com/openlawlibrary/taf/pull/559/commits/c317b2caaa6c194fb295099cd3530beb5958507d). I also tried to resolve most of the failing tests. 

It seems like only a single test is now failing. If these changes looks good, could we please take a look and see if the test can pass? Then I think we can merge!

renatav:
@n-dusan I don't think that the last validated commit is unnecessary. It is updated when the updater is run without the `--exclude-targets` option. You could update the auth repo without updating any of the targets. It's the last validated commit of that set of repositories. While it could be possible to implement the updater without it, I think that that is still useful to keep track of.

n-dusan:
>While it could be possible to implement the updater without it, I think that that is still useful to keep track of.

Thanks for the clarification. That makes sense now. Could you expand on the reasons why keeping track of it is useful? It would be helpful to add an explanation in the documentation and include comments in the code so that its purpose is clear to others in the future. 

Without those, it might be interpreted as unnecessary or overlooked. Which then leads to confusion.

renatav:
@n-dusan I updated the docstrings, ready for another review

n-dusan (approved):

PR #580, opened 2025-01-16
feat: add `no_verify` to methods that push

n-dusan:

Sometimes we don't want to run pre-push hook validations when calling out to git push from our programs. To avoid running any git hooks, we bypass with `--no-verify`.


renatav (commented):
I am not sure that I like the idea to allow bypassing verification in this way. At least not directly in taf. I'd be more comfortable with defining this outside of TAF

renatav (approved):
After thinking about this some more, this should be harmless enough. TAF's API functions will never use the flag

PR #597, opened 2025-03-04
Performance tests

### LINKED ISSUE #571: Implement performance regression updater tests

renatav:

Integrate benchmark tests into CI:
- We now run benchmark tests after every push. Made it possible to use a JSON configuration file to set different performance thresholds when necessary. This is useful especially when new features are added that might slow down the system because they do more work. The default value is set to 10.
- Every run compares the current results with the master branch's performance. This meant that some custom download code was necessary, as we want to download from a different branch.
- This custom JSON file also helps us keep track of why and when we needed to change these thresholds
- When changes are merged into the master branch, we generate new benchmark results and upload the corresponding artifacts. 
- Add scheduled builds set to run every 90 days to prevent our artifacts from being automatically deleted due to GitHub's cleanup policies.

There are probably many ways in which we can compare this whole logic, but I think this is a good starting point and we'll have some basic performance tests. I think that we can close the issue and open new ones if we think of any improvements

Closes #571


docs/developers/performance-regression-tests.md
```diff
@@ -0,0 +1,39 @@
...
+
+The benchmark threshold can be defined for each branch to accommodate specific performance criteria associated with different types of work (e.g., feature development, bug fixes). This threshold is set in the `benchmark_thresholds.json` file. If not set, threshold is set to 10 by default.
+
+### Example `benchmark_thresholds.json`:
+
+```json
+{
+  "feature/cool-feature": 20,
+  "bugfix/urgent-fix": 15
+}
```
n-dusan, line 32: suggestion: A sentence or two explaining what the threshold number means would be useful

benchmark_thresholds.json
```diff
@@ -0,0 +1,3 @@
+{
+    "renatav/performance-tests": 15
```
n-dusan, line 2: question: can we bring this threshold to be lower? Is 15 times slower before it errors out too much?

.github/workflows/ci.yml
```diff
@@ -31,6 +31,13 @@ jobs:
     strategy:
       matrix:
         python-version: ${{ fromJSON(needs.set_python_versions.outputs.all_versions) }}
+
+    env:
+      ARTIFACT_NAME: benchmark-results${{ matrix.python-version }}
+      BRANCH_NAME: 'master' # needs to be changed to master before merging
```
n-dusan, line 37:
```suggestion
      BRANCH_NAME: 'master'
```

.github/workflows/ci.yml
```diff
@@ -61,6 +68,111 @@ jobs:
...
+              }
+            }
+
+            if (latestArtifact) {
+              console.log(`Found latest artifact: ${latestArtifact.id}`);
+              core.setOutput('artifact_id', latestArtifact.id.toString());
+            } else {
+              console.log('No matching artifacts found.');
+              core.setOutput('artifact_id', '');
+            }
```
n-dusan, line 118: question: could we check in this javascript (and other) code somewhere in the `.github` (e.g. `.github/scripts`) directory and have GitHub actions require the file? The benefit is that we have syntax highlighting and would be easier to debug [1]?

[1] - https://github.com/orgs/community/discussions/26248

n-dusan (commented):
Looks awesome. Left some minor suggestions

docs/developers/performance-regression-tests.md
renatav: I removed this file and merged everything with the testing docs that I asked Cal to write

benchmark_thresholds.json
renatav: That's 15%.  The default is 10%, this is just an example of how to overwrite the default value

.github/workflows/ci.yml
```diff
@@ -56,10 +63,69 @@ jobs:
           git config --global user.name oll-bot
           git config --global user.email developers@openlawlib.org
 
-      - name: Run pre-commit and test with pytest
+      # - name: Run pre-commit and test with pytest
+      #   run: |
+      #     pre-commit run --all-files
+      #     pytest taf/tests
+
```
n-dusan, line 70: question: is running `pytest taf/tests/` and pre-commit supposed to be commented out?

.github/workflows/ci.yml
```diff
@@ -56,10 +63,69 @@ jobs:
...
+        if: steps.get_artifact.outputs.artifact_id == '' || steps.get_artifact.outputs.artifact_id == null
         run: |
-          pre-commit run --all-files
-          pytest taf/tests
+          python -m pytest --benchmark-only --benchmark-json=0001_output.json
+
+      # Use a step to extract the branch name from github.ref
+      - name: Extract Branch Name
+        id: extract_branch
+        run: echo "::set-output name=branch::$(echo ${GITHUB_REF#refs/heads/})"
```
n-dusan, line 121: suggestion: set-output is getting deprecated. Could we please use the new approach [1]? The deprecation notice is [here](https://github.com/openlawlibrary/taf/actions/runs/13805123753). 

[1] - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

docs/developers/benchmark-testing.md
```diff
@@ -20,5 +20,38 @@ If you want to cause failure in the case of regression, also add "–-benchmark-
...
+```json
+{
+  "feature/cool-feature": 20,
+  "bugfix/urgent-fix": 15
+}
+```
+
+## Artifact Handling
+
+- **Uploading Artifacts**: After benchmarks are run, the resulting `0001_output.json` file is uploaded as an artifact to GitHub Actions, allowing it to be used in future benchmark comparisons.
```
n-dusan, line 56: suggestion: adding link to the already uploaded artifacts (from some older GitHub Actions build as an example) would be very useful here

PR #694, opened 2026-01-20
Add Python expression filtering for repository exclusion

renatav:
## Summary
Adds `--exclude-filter` option to exclude repositories based on their custom data using Python expressions.

Previously, repositories could only be excluded via glob patterns (`--exclude-target`). This works for excluding specific repository names but doesn't allow filtering based on repository metadata like type and other custom attributes.

Considered using jq for filtering, but it's an external binary that complicates Windows deployments. Python libraries like jmespath are either unmaintained or don't support Python 3.13+.

- New `filter_expr` parameter accepts Python expressions where `repo` represents the repository's custom data
- AST validation ensures only safe operations (comparisons, boolean logic, dict access) are allowed
- Restricted eval namespace (`{"__builtins__": {}}`) prevents dangerous code execution
- Returns list of repository names matching the filter, which are then excluded from update

## Usage Examples
```bash
# Exclude HTML repositories
taf repo clone  git@github.com:oll-test-repos/cityofsanmateo-law.git --exclude-filter "repo['type'] == 'html'"

# Exclude repos serving latest
taf repo clone  git@github.com:oll-test-repos/cityofsanmateo-law.git --exclude-filter "repo.get('serve') == 'latest'"

# Exclude HTML and XML repositories
taf repo clone - git@github.com:oll-test-repos/cityofsanmateo-law.git -exclude-filter "repo['type'] in ('html', 'xml')"
```

Also updated exclusion logic:
- Previously, update required the same `--exclude-target` parameters on every run to continue excluding repositories
- Now, once if a repository is excluded during clone, it remains excluded on subsequent updates unless `--sync-all` is specified
- Exclusion state is tracked via `last_validated_commit`
- `--sync-all` forces synchronization of all repositories, ignoring previous exclusions


n-dusan (approved):
Looks good, thanks!

PR #719, opened 2026-04-01
Updater: change excluded repos handling

### LINKED ISSUE #716: Updater: change excluded target repos handling
Exclusion of target repositories should not be handled by the updater in `update` operation. Instead, `only clone` operation will have the option to exclude certain repo/repos.
- [ ] Modify contents of `last_validated_commit` file in the following way:
1. Put last validated commit of the excluded repo to None
2. Add `excluded` entry with exclusion filter or excluded repos

Example for excluded archived html repo:
```
{
    "DCCouncil/law-docs": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "DCCouncil/law-xml": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "DCCouncil/law-html": None,
    "DCCouncil/law-html-2026": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "DCCouncil/law-xml-codified": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "DCCouncil/law-static-assets": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "DCCouncil/law-rdf": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "DCCouncil/law": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "last_validated_commit": "80dbb7329f5eca34d04fa65b38a1cf6ced3e7311",
    "exclude_filter": "repo["archived"] is not None"
}
```

This means that update and reset operation should be extended to check for excluded entries in last validated commit file and use that info to prevent potential failures.

- [ ] Remove `--exclude-target` flag:
Extend `--exclude-filter` to handle this functionality instead. If `repo['name']` is configured in the filter, the filtering should be done against keys in `repositories.json`.

sale3:

Target repo exclusion is set only in clone operation and the pattern is stored in `last_validated_commit` file.
Excluded target repos' last validated commits are set to `None`.
`--exclude-target` is removed in favor of `--exclude-filter`.
`--exclude-filter` is extended to filter repos by name to match `--exclude-target` implementation.

In `auth_repo.py`:
Removed excluded repos handling for methods `targets_data_by_auth_commit` and `sorted_commits_and_branches_per_repositories` since it is not used anywhere. Introducing `--exclude-filter` would lead to coupling with repositoriesdb.

Tests:
Removed `test_update_after_update_with_exclude`, `test_update_after_update_with_exclude_with_invalid_commit` and `test_full_update_after_partial_update` since update with exclude is no longer feasible.
Added verification for expected `last_validated_commit` entires - `verify_excluded_lvc_entries` function.

Closes: #716 

### DIFF
--- a/taf/auth_repo.py
+++ b/taf/auth_repo.py
@@ -417,7 +417,6 @@ class AuthenticationRepository(GitRepository):
         commits: List[Commitish],
         target_repos: Dict[str, GitRepository],
         custom_fns: Optional[Dict[str, Callable]] = None,
-        excluded_target_globs: Optional[List[str]] = None,
         last_commits_per_repos: Optional[Dict[Commitish, List]] = None,
     ) -> Dict[str, Dict[Commitish, Dict[str, Any]]]:
         """
@@ -443,14 +442,8 @@ class AuthenticationRepository(GitRepository):
             target_repos=target_repos,
             last_commits_per_repos=last_commits_per_repos,
         )
-        excluded_target_globs = excluded_target_globs or []
         for commit in commits:
             for target_path, target_data in targets[commit].items():
-                if any(
-                    fnmatch.fnmatch(target_path, excluded_target_glob)
-                    for excluded_target_glob in excluded_target_globs
-                ):
-                    continue
 
                 target_branch = target_data.get("branch")
                 target_commit = target_data.get("commit")
@@ -474,7 +467,6 @@ class AuthenticationRepository(GitRepository):
         commits: List[Commitish],
         target_repos: Optional[Dict[str, GitRepository]] = None,
         custom_fns: Optional[Dict[str, Callable]] = None,
-        excluded_target_globs: Optional[List[str]] = None,
     ) -> Dict[str, Dict[str, List[Dict[str, Any]]]]:
         """Return a dictionary consisting of branches and commits belonging
         to it for every target repository:
@@ -505,18 +497,8 @@ class AuthenticationRepository(GitRepository):
         repositories_commits: Dict = defaultdict(dict)
         targets = self.targets_at_revisions(commits, target_repos=target_repos)
         previous_commits: Dict = {}
-        skipped_targets = []
-        excluded_target_globs = excluded_target_globs or []
         for commit in commits:
             for target_path, target_data in targets[commit].items():
-                if target_path in skipped_targets:
-                    continue
-                if any(
-                    fnmatch.fnmatch(target_path, excluded_target_glob)
-                    for excluded_target_glob in excluded_target_globs
-                ):
-                    skipped_targets.append(target_path)
-                    continue
                 target_branch = target_data.get("branch")
                 target_commit = target_data.get("commit")
                 previous_data = previous_commits.get(target_path)
--- a/taf/tests/test_api/repo/api/test_create_repository.py
+++ b/taf/tests/test_api/repo/api/test_create_repository.py
@@ -128,5 +128,4 @@ def test_create_repository_when_add_repositories_json(
     for role in ("targets", "delegated_role", "inner_role"):
         assert role in targets_roles
     check_if_targets_signed(auth_repo, "targets", "repositories.json", "mirrors.json")
-    # we are not validating target repositories, just the authentication repository
     validate_repository(repo_path)
--- a/taf/tests/test_api/repo/api/test_create_repository.py
+++ b/taf/tests/test_api/repo/api/test_create_repository.py
@@ -129,4 +129,4 @@ def test_create_repository_when_add_repositories_json(
         assert role in targets_roles
     check_if_targets_signed(auth_repo, "targets", "repositories.json", "mirrors.json")
     # we are not validating target repositories, just the authentication repository
-    validate_repository(repo_path, excluded_target_globs="*")
+    validate_repository(repo_path)
--- a/taf/tests/test_updater/update_utils.py
+++ b/taf/tests/test_updater/update_utils.py
@@ -17,22 +17,22 @@ from taf.updater.updater import UpdateConfig, clone_repository, update_repositor
 
 
 def check_last_validated_commit(
-    clients_auth_repo_path, all_target_repositories=None, excluded_targets=None
+    clients_auth_repo_path, all_target_repositories=None, excluded_target_names=None
 ):
     # check if last validated commit is created and the saved commit is correct
     client_auth_repo = AuthenticationRepository(path=clients_auth_repo_path)
     head_sha = client_auth_repo.head_commit()
     last_validated_data = client_auth_repo.last_validated_data
     assert last_validated_data[client_auth_repo.name] == head_sha.value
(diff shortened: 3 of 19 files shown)


taf/tests/test_api/repo/api/test_create_repository.py
```diff
@@ -129,4 +129,4 @@ def test_create_repository_when_add_repositories_json(
         assert role in targets_roles
     check_if_targets_signed(auth_repo, "targets", "repositories.json", "mirrors.json")
     # we are not validating target repositories, just the authentication repository
-    validate_repository(repo_path, excluded_target_globs="*")
+    validate_repository(repo_path)
```
renatav, line 131: With the removal of excluded_target_globs, we should be validating the target repositories too, so the comment above is no longer correct.I am not sure why we were skipping targets, maybe because validation of targets was failing when this was first added? I think we should just remove the comment

taf/tests/test_updater/update_utils.py
```diff
@@ -33,7 +32,7 @@ def check_last_validated_commit(
 
     if all_target_repositories and excluded_targets:
         for target_repo in all_target_repositories:
-            if target_repo not in excluded_targets:
+            if target_repo.name not in excluded_targets:
                 assert last_validated_data[target_repo.name] == head_sha.value
             else:
                 assert last_validated_data.get(target_repo.name) != head_sha.value
```
renatav, line 38: If I am understanding the requirement correctly, an excluded target should never appear in last validated data? So should we check `target_repo.name not in last_validate_data` here?

taf/tests/test_updater/update_utils.py
```diff
@@ -33,7 +32,7 @@ def check_last_validated_commit(
 
     if all_target_repositories and excluded_targets:
         for target_repo in all_target_repositories:
-            if target_repo not in excluded_targets:
```
renatav, line 36: If we are passing repo names to this function now, let's call rename the parameter to `excluded _target_names` for clarity

taf/tests/test_updater/update_utils.py
```diff
@@ -416,6 +420,22 @@ def verify_repos_exist(
...
 
 
+def verify_excluded_lvc_entries(
+    client_dir: Path, origin_auth_repo: AuthenticationRepository, excluded: list
+):
+    client_auth_repo = AuthenticationRepository(path=client_dir / origin_auth_repo.name)
+    last_validated_data = client_auth_repo.last_validated_data
+    assert last_validated_data is not None
+    if last_validated_data.get("exclude_filter") is not None:
+        assert last_validated_data["exclude_filter"]
```
renatav, line 430: This assert does not really check anything given the if condition. Maybe `if excluded`:    assert "exclude_filter" in last_validated_data` else assert  exclude_filter" not in last_validated_data` would be more useful.

taf/updater/updater_pipeline.py
```diff
@@ -1111,39 +1115,59 @@ def set_auth_commit_for_target_repos(self):
...
+            excluded_repo_names = repositoriesdb.get_repository_names_by_expression(
+                self.state.users_auth_repo, filter_expr=self.exclude_filter
+            )
+            if excluded_repo_names:
+                self.excluded_target_globs.extend(excluded_repo_names)
+
+        last_validated_data = self.state.users_auth_repo.last_validated_data
+        if last_validated_data:
+            if len(last_validated_data) == 1:
+                # check if old last validated format
```
renatav: Before we started saving more information in `last_validated_commit`, that used to be a file containing only the last validate auth repo commit. It wasn't a json file at all and we really don't know which repositories the user excluded while cloning. Same if they excluded some repositories before we started saving that information. I don't think there is a point in trying to this load json here. What should we do if the user deleted this `last_validated_commit` file? or modified it manually?

taf/updater/updater_pipeline.py
```diff
@@ -1111,39 +1115,59 @@ def set_auth_commit_for_target_repos(self):
...
-                    self.excluded_target_globs.extend(skipped_repositories)
+                try:
+                    json.loads(last_validated_file_content)
+                except json.decoder.JSONDecodeError:
+                    return UpdateStatus.SUCCESS
+
+            all_repositories = repositoriesdb.load_repositories_json(
+                self.state.users_auth_repo
+            )
+            if all_repositories is not None:
```
renatav: This was my old way of handling this, so without saving `exclude_filter`. Just look for repositories not in `last_validated_data` and skip them. If are going to start storing `exclude_filter`, we should either validate rely on it primarily and just check if repositories not in last validated data and not on disk match the filter? What if there is a discrepancy here? Do we even need `excluded_target_globs` if we have `exclude_filter`? I added that automatic detection of excluded target repositories since there was no way to read this information.

taf/auth_repo.py
```diff
@@ -417,7 +417,6 @@ def targets_data_by_auth_commits(
         commits: List[Commitish],
         target_repos: Dict[str, GitRepository],
         custom_fns: Optional[Dict[str, Callable]] = None,
-        excluded_target_globs: Optional[List[str]] = None,
```
renatav, line 420: Shouldn't we replace this with the filter?

taf/tests/test_updater/update_utils.py
sale3: mypy was complaining if there was no null check, so that's why I put it like that.

taf/updater/updater_pipeline.py
sale3: Agreed to remove support for old format.
Added test case that handles deletion of `last_validated_commit` file.

taf/updater/updater_pipeline.py
sale3: I just realized that we can get rid of this completely. I think that it's not likely to happen that `repositories.json` doesn't have one of the target repos as an entry?

taf/auth_repo.py
sale3: Removed it since it's unused, I documented that in the PR description.

taf/updater/updater_pipeline.py
renatav: It very unlikely, but I wouldn't make that assumption.

taf/auth_repo.py
renatav: Sounds good. Sorry, I didn't read the description carefully as I assumed I knew what the PR was about

taf/tests/test_updater/update_utils.py
sale3: The repo should appear, but it's entry should be `None`.

renatav (approved):
Tested and works well. Found a minor edge-case, might not even be addressing now. When you clone with an `exclude-filter` and then remove the filter line from the `last_validated_commit` file and run update, it clones the missing repository but prints All repositories are up-to-date with no changes or errors.

PR #724, opened 2026-04-16
Fix: enable exclude filter for validation, rework partial update detection

sale3:

Reworks partial-update detection for exclude-filter flows. Updates with excluded targets are no longer treated as partial when all included repositories validate successfully, and last_validated_commit is now persisted for those runs.

Tested the following:
- Cloned with `repo clone --exclude-filter "repo['type']=='html'" git@github.com:oll-test-repos/mohicanlaw-law.git`
- Ran `taf repo update` and confirmed that the html repo was not cloned
- Deleted the `exclude-filter` entry from `last_validated_commit` and confirmed that the repo was cloned as expected
- Deleted `last_validated_commit` and ran update, condifrming that the validation started from the beginning and was completed successfully.
- Ran integration tests

~Docstrings have been included and/or updated, as appropriate~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Fixed
 
+- Enable `--exclude-filter` for validation ([724])
+
+[724]: https://github.com/openlawlibrary/taf/pull/724
+
 ## [0.38.0]
 
 ### Added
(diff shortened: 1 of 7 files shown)


n-dusan (approved):
question (non-blocking): is it worth adding tests for exclude-filter scenarios to reduce the need for manual testing, especially once codify rolls out?

n-dusan (approved):

PR #732, opened 2026-05-05
fix: raise PygitError instead of AttributeError when pygit2 unavailable

### LINKED ISSUE #728: Crash with AttributeError when pygit2 is unavailable
When pygit2 is not installed or fails to initialize, several methods in taf/git.py crash with opaque AttributeError messages rather than a meaningful error.

Example:
AttributeError: 'NoneType' object has no attribute 'branches'. The root cause is that pygit_repo can return None or raise PygitError, but many callers do not handle this case.

Affected methods (non-exhaustive):

all_commits_on_branch — calls repo.branches.get(branch) without None check
branches — calls repo.branches.local / repo.branches.remote without None check
is_detached_head — calls repo.head_is_detached without None check
list_pygit_commits — calls repo.branches.get(branch) without None check
branch_exists — calls repo.branches.get(branch_name) without None check
is_bare_repository — calls self.pygit_repo.is_bare without None check
Expected behavior:

If pygit2 is not available or fails to initialize, a clear and actionable error should be raised.

SAY-5:

Fixes #728. Instead of catching `AttributeError` in individual repository accessors, this now fails at the `pygit2` integration boundary: `git.py` sets a `PYGIT2_AVAILABLE` flag during import, `pygit` raises `PygitError("pygit2 is not installed")` before touching the repository when that dependency is unavailable, and `clone_from_disk` applies the same guard before calling `pygit2` directly. Regression tests cover both guarded paths.

### DIFF
--- a/taf/git.py
+++ b/taf/git.py
@@ -174,7 +174,12 @@ class GitRepository:
     @property
     def is_detached_head(self) -> bool:
         repo = self.pygit_repo
-        return repo.head_is_detached
+        try:
+            return repo.head_is_detached
+        except AttributeError:
+            raise PygitError(
+                "pygit2 repository is unavailable; cannot determine HEAD state"
+            )
 
     @property
     def is_git_repository(self) -> bool:
@@ -238,11 +243,11 @@ class GitRepository:
     @property
     def is_bare_repository(self) -> bool:
         if self._is_bare_repo is None:
-            if self.pygit_repo is not None:
+            try:
                 self._is_bare_repo = self.pygit_repo.is_bare
-            else:
-                raise GitError(
-                    "Cannot determine if repository is a bare repository. Cannot instantiate pygit repository"
+            except AttributeError:
+                raise PygitError(
+                    "pygit2 repository is unavailable; cannot determine if repository is bare"
                 )
 
         return self._is_bare_repo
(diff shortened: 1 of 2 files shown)


renatav:
@SAY-5 Thank you for your contribution! Could we ask you to sign our [Contributor License Agreement](https://cla-assistant.io/openlawlibrary/taf?pullRequest=732)?

SAY-5:
I have signed the CLA @renatav

renatav (commented):
Thanks for signing the agreement! This issue is actually a bit tricky as it requires understanding how pygit2 is wired through the codebase.

This error showed up on CI when an unrelated problem led to a partial installation. I think the best fix is to catch this at the instantiation level rather than updating individual methods:

- git.py (top-level): Wrap import pygit2 and from .pygit import PyGitRepository in a try/except ImportError and set a PYGIT2_AVAILABLE flag.
- pygit property: Check PYGIT2_AVAILABLE first and raise PygitError("pygit2 is not installed") before doing anything else. 
- clone_from_disk: This one also needs a PYGIT2_AVAILABLE guard since it calls pygit2.clone_repository directly without going through pygit_repo.

One thing worth knowing: PygitError inherits from TAFError, not GitError, so it will propagate past except GitError handlers. That's fine as long as the error message is descriptive.

renatav (commented):
This looks good now! Thanks for your contribution and interest in the project.

renatav (approved):

PR #740, opened 2026-05-28
fix: out-of-band fixes

n-dusan:

1. Fall back to the per-repo key when the global last_validated_commit key is missing from the validated-commit file (written without it by older TAF on partial updates).
2. Drop deps absent from the latest commit during deduplication so a moved dependency doesn't get validated against its old out of band hash from the originating repo's history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Fixed
 
+- Fix false out-of-band authentication failures on update and clone
+
 
 ## [0.38.3]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Fixed
 
-- Fix false out-of-band authentication failures on update and clone
+- Fix false out-of-band authentication failures on update and clone ([740])
+
+[740]: https://github.com/openlawlibrary/taf/pull/740
 
 
 ## [0.38.3]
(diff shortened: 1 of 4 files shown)


taf/repositoriesdb.py
```diff
@@ -732,6 +734,13 @@ def _get_deduplicated_target_or_auth_repositories(
             # will overwrite older repo with newer
             repositories[name] = repo
 
+    # Drop repos absent from the latest commit so stale historical entries
+    # don't survive after a dependency is removed or moved.
+    latest_repos = set(all_repositories.get(commits[-1], {}).keys())
```
renatav: I think this issue should be resolved elsewhere. This was designed to simply return previously loaded repositories, given a list of commits. So the caller should only provide the newest commit - or maybe we don't even want to load repos associated with older commits. Is this in the pipeline?

taf/updater/updater_pipeline.py
```diff
@@ -1167,9 +1167,14 @@ def validate_out_of_band_and_update_type(self):
             f"{self.state.auth_repo_name}: Validating out of band commit and update type..."
         )
         try:
+            # Fall back to the per-repo key in case the global
+            # "last_validated_commit" key was not written (older TAF bug).
+            _prior_lvc = self.state.users_auth_repo.last_validated_commit or (
+                self.state.last_validated_data or {}
```
renatav: We should move this to `last_validated_commit` since it is already loading from `last_validated_data`

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning][semver].
 
 ### Fixed
 
+- Fix false out-of-band authentication failures on update and clone
```
renatav: Missing PR link

renatav (approved):

PR #743, opened 2026-06-04
refact: Performance improvements

### LINKED ISSUE #709: Updater: transition `auth repo clone to temp` to local clone
In the new updater pipeline there is a stage where all local repositories are consistent (in sync with last validated commit). If auth repo in this state is not up to date with remote, usual pipeline steps are applied.
One of these steps is cloning of auth repo to temp, which is done from remote. In order to improve performance, this clone should be done locally (as it is done for target repos) since all repositories are in consistent state.

n-dusan:

Passthrough on performance benefits by cutting down network time. 

- taf repo clone is 50% faster. We skip the redundant network fetch when materializing fresh target repos. Validated objects are already in the temp clone, so origin is never re-fetched; materialization is also parallelized.
- taf repo update is 70% faster: build the temp validation auth repo from the user's on-disk copy + a delta fetch, and refresh the user's auth repo from the temp clone. This is cutting two full network pulls of auth history down to one incremental one.
- Target repo network fetch de-duplicated on update: clone_target_repositories_to_temp and get_target_repositories_commits previously each hit origin independently; they now share a single fetch.
- Hot-path debug logging removed: ~125k eager f-string debug() calls inside _get_blob_at_path (invoked ~63k per clone) are gone, saving ~4–5s of pure logging overhead.

Closes #709

### DIFF
--- a/taf/git.py
+++ b/taf/git.py
@@ -179,6 +179,8 @@ class GitRepository:
 
     _is_bare_repo = None
 
+    _is_git_repository = None
+
     @property
     def remotes(self) -> List[str]:
         if self._remotes is None:
@@ -195,17 +197,44 @@ class GitRepository:
 
     @property
     def is_git_repository(self) -> bool:
-        """Check if the given path is the root of a Git repository."""
+        """Check if the given path is inside a Git repository (work tree or bare)."""
         # This is used when instantiating a PyGitRepository repo, so do not use
-        # it here
+        # self.pygit / self.pygit_repo here (would recurse via the pygit property).
+        #
+        # Only a positive result is cached. A path that is not (yet) a repository
+        # can become one later - e.g. a GitRepository instance is created for a
+        # path before init_repo()/clone() runs, or before the updater (a
+        # different instance, or an external process) materializes it on disk.
+        # Caching the negative result would make this instance permanently report
+        # the path as a non-repository even after a valid .git appears, so we
+        # recompute until the answer is True.
+        if self._is_git_repository:
+            return True
+        # pygit2.discover_repository is a module-level function that searches
+        # upward for a repository, matching the semantics of the subprocess
+        # `rev-parse --is-inside-work-tree` / `--is-bare-repository` checks but
+        # without spawning a process (~150-400ms each on Windows).
+        if PYGIT2_AVAILABLE and not self.allow_unsafe:
+            try:
+                if pygit2.discover_repository(str(self.path)) is not None:
+                    self._is_git_repository = True
+                    return True
+                return False
+            except Exception:
+                # fall back to the subprocess check (e.g. ownership errors)
+                pass
+        # subprocess fallback: pygit2 unavailable, or allow_unsafe is set
+        # (libgit2 does not honor the `-c safe.directory` override the
+        # subprocess path uses)
         try:
             result = self._git("rev-parse --is-inside-work-tree", reraise_error=True)
             if result == "true":
+                self._is_git_repository = True
                 return True
             result = self._git("rev-parse --is-bare-repository", reraise_error=True)
             if result == "true":
+                self._is_git_repository = True
                 return True
-
         except GitError:
             return False
         return False
@@ -213,8 +242,6 @@ class GitRepository:
     @property
     def is_git_repository_root(self) -> bool:
         try:
-            if not self.is_git_repository:
-                return False
             repo = self.pygit_repo
             if repo is None:
                 return False
@@ -225,7 +252,7 @@ class GitRepository:
                 return Path(repo.path).resolve() == git_path.resolve() and (
                     git_path.is_dir() or git_path.is_file()
                 )
-        except PygitError:
+        except (PygitError, GitError):
             return False
 
     @property
@@ -321,7 +348,16 @@ class GitRepository:
         In those cases, get the default branch from `git remote show origin`.
 
         If the repository does not have a remote set, use `symbolic-ref` HEAD
+
+        The cheap symbolic-ref reads (steps 1 and 3) are done in-process via
+        pygit2 when possible to avoid spawning subprocesses (~150-400ms each on
+        Windows); each falls back to the subprocess equivalent on any failure,
+        so behavior is a strict superset of the previous implementation.
         """
+        # step 1: refs/remotes/origin/HEAD
+        branch = self._symbolic_ref_branch_via_pygit("refs/remotes/origin/HEAD")
+        if branch is not None:
+            return branch
         try:
             branch = self._git(
                 "symbolic-ref refs/remotes/origin/HEAD --short", reraise_error=True
@@ -330,6 +366,7 @@ class GitRepository:
         except GitError as e:
             self._log_debug(f"Could not get remote HEAD at {self.path}: {e}")
             pass
+        # step 2: git remote show origin (may contact the network; subprocess only)
         try:
             result = self._git("remote show origin", reraise_error=True)
             match = re.search(r"HEAD branch:(.*)", result)
@@ -340,6 +377,10 @@ class GitRepository:
                 f"Could not get HEAD branch with git remote show origin at {self.path}: {e}"
             )
             pass
+        # step 3: HEAD
+        branch = self._symbolic_ref_branch_via_pygit("HEAD")
+        if branch is not None:
+            return branch
         try:
             return self._git("symbolic-ref HEAD --short", reraise_error=True)
         except GitError as e:
@@ -349,6 +390,45 @@ class GitRepository:
             self,
             message="Could not determine default branch from local repository",
         )
+
+    def _symbolic_ref_branch_via_pygit(self, ref_name: str) -> Optional[str]:
+        """Return the short branch name a symbolic ref points to, read in-process
+        via pygit2, or None on any failure (so the caller falls back to the
+        subprocess equivalent). Only symbolic refs yield a branch name.
+
+        Opens a throwaway pygit2.Repository instead of using self.pygit_repo:
+        this method can run before the repository exists on disk (e.g. during
+        construction, ahead of init_repo()/clone()), and discover_repository
+        walks upward, so a not-yet-initialized path nested inside another
+        repository (e.g. a test fixture under this checkout) would otherwise
+        resolve to - and permanently cache on self - that unrelated parent
+        repository.
+        """
+        if not PYGIT2_AVAILABLE:
+            return None
+        try:
+            discovered = pygit2.discover_repository(str(self.path))
+            if discovered is None:
+                return None
+            discovered_path = Path(discovered).resolve()
+            self_path = self.path.resolve()
+            if (
+                discovered_path != self_path
+                and self_path not in discovered_path.parents
+            ):
+                # discovered a different (e.g. parent) repository
+                return None
+            repo = pygit2.Repository(discovered)
+            ref = repo.references.get(ref_name)
+            if ref is None or ref.type != pygit2.GIT_REF_SYMBOLIC:
+                return None
+            target = ref.target  # e.g. "refs/remotes/origin/main" or "refs/heads/main"
+            for prefix in ("refs/remotes/origin/", "refs/heads/"):
+                if target.startswith(prefix):
+                    return target[len(prefix) :]
+            return target
+        except Exception:
+            return None
 
     def _get_default_branch_from_remote(self, url: str) -> str:
         if not self.is_git_repository:
@@ -783,6 +863,10 @@ class GitRepository:
         if not cloned:
             self.raise_git_access_error(CloneRepoException)
 
+        # the path is now a repository; drop any cached negative result from
+        # before the clone
+        self._is_git_repository = None
+
         if self.default_branch is None:
             self.default_branch = self._determine_default_branch()
 
@@ -799,6 +883,9 @@ class GitRepository:
             raise PygitError("pygit2 is not installed")
         self.path.mkdir(parents=True, exist_ok=True)
         pygit2.clone_repository(local_path, self.path, bare=is_bare)
+        # the path is now a repository; drop any cached negative result from
+        # before the clone
+        self._is_git_repository = None
         if not self.is_git_repository:
             raise GitError(
                 self, message=f"Could not clone repository from local path {local_path}"
@@ -1438,6 +1525,7 @@ class GitRepository:
             self.path.mkdir(exist_ok=True, parents=True)
         flag = "--bare" if bare else ""
         self._git(f"init {flag}", error_if_not_exists=False)
+        self._is_git_repository = None
         if self.urls is not None and len(self.urls):
             self._git("remote add origin {}", self.urls[0])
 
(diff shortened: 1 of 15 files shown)


taf/updater/handlers.py
```diff
@@ -125,6 +130,18 @@ def _fetch(self, url):
             )
             raise e
 
+    def download_bytes(self, url: str, max_length: int) -> bytes:
+        # the default FetcherInterface implementation round-trips the data
```
renatav: I don't fully understand from this docstring when the rest of the validation takes place and why this check is needed and the others are not

taf/updater/in_memory_updater.py
```diff
@@ -0,0 +1,41 @@
...
+            f"python-tuf internals changed: Updater.{_method} no longer exists. "
+            "InMemoryUpdater must be updated to match the installed tuf version."
+        )
+
+
+class InMemoryUpdater(Updater):
+    """TUF Updater whose local metadata cache is an in-memory dict.
+
+    TAF runs one Updater per validated commit, so the "local metadata
+    directory" is written and re-read thousands of times per update. Keeping
```
renatav: I think we need a better explanation of what this metadata cache and how this sharing works. Metadata files change all the time, or we need to validate that they stayed the same. Could you clarify what this is doing and why it's safe to use it? Are we confident that the test is covering everything?

taf/updater/updater_pipeline.py
```diff
@@ -899,8 +901,62 @@ def clone_auth_to_temp(self):
...
+                    else self.state.users_auth_repo.get_remote_url()
+                )
+                if not remote_url:
+                    raise UpdateFailedError(
+                        "URL cannot be determined. Please specify it"
+                    )
+                self.state.validation_auth_repo.clone_bare_from_local(
+                    self.state.users_auth_repo.path
+                )
+                self.state.validation_auth_repo._git(
```
renatav: I don't think we should we calling `_.git` anywhere in this pipeline directly. Let's move all of these calls to `git.py`, add docstrings (and tests if that makes sense). Let's also check if there is any code duplication or something that can be solved by reusing an existing method, but with e.g. an additional input parameter.

taf/updater/updater_pipeline.py
```diff
@@ -899,8 +901,62 @@ def clone_auth_to_temp(self):
             self.state.validation_auth_repo = AuthenticationRepository(
                 path=path, urls=self.urls, alias="Validation repository"
             )
-            self.state.validation_auth_repo.clone(bare=True)
-            self.state.validation_auth_repo.fetch(fetch_all=True)
+            if self.state.existing_repo and self.state.users_auth_repo:
+                # Seed the temp validation repo from the user's on-disk copy (fast,
```
renatav: I think it would be better to move this whole thing to a separate method (can be a helper in this module imo)

taf/updater/updater_pipeline.py
```diff
@@ -899,8 +901,62 @@ def clone_auth_to_temp(self):
...
+                    remote_url,
+                    log_error=True,
+                    reraise_error=True,
+                )
+                self.state.validation_auth_repo._git(
+                    "fetch origin +refs/heads/*:refs/heads/*",
+                    log_error=True,
+                    reraise_error=True,
+                )
+                # Determine default_branch from the real remote, not from the
```
renatav: I'm finding these comments to be overly verbose. E.g. The remote's HEAD is the authoritative default branch (matches the clean-clone path) sounds good enough here

taf/updater/updater_pipeline.py
```diff
@@ -899,8 +901,62 @@ def clone_auth_to_temp(self):
...
+                    self.state.validation_auth_repo.get_default_branch(remote_url)
+                )
+                # seed the cache with the authoritative remote value so later
+                # detections for this path don't shell out (clone_bare_from_local
+                # cleared it on purpose; this re-populates it correctly)
+                if self.state.validation_auth_repo.default_branch:
+                    _default_branch_cache[str(self.state.validation_auth_repo.path)] = (
+                        self.state.validation_auth_repo.default_branch
+                    )
+                if self.state.validation_auth_repo.default_branch:
```
renatav: Same if condition appearing twice `if self.state.validation_auth_repo.default_branch`

taf/updater/updater_pipeline.py
```diff
@@ -1240,10 +1303,38 @@ def clone_or_fetch_users_auth_repo(self):
             f"{self.state.auth_repo_name}: Cloning or updating user's authentication repository..."
         )
         try:
+            temp_path = self.state.validation_auth_repo.path
             if self.state.existing_repo:
-                self.state.users_auth_repo.fetch(fetch_all=True)
+                # Fetch new commits from the already-fetched temp validation repo
+                # (local, no network) instead of re-fetching from the real remote.
+                self.state.users_auth_repo._git(
```
renatav: This should definitely be a separate method. The comment preceding it can be added to the docstring

taf/updater/updater_pipeline.py
```diff
@@ -1240,10 +1303,38 @@ def clone_or_fetch_users_auth_repo(self):
...
             else:
-                self.state.users_auth_repo.clone(bare=self.bare)
+                # Materialize the user's auth repo from the validated temp rather
+                # than cloning from the network a second time.
+                remote_url = (
+                    self.state.users_auth_repo.urls[0]
+                    if self.state.users_auth_repo.urls
+                    else None
+                )
+                # Pass the default branch so clone_from_disk calls set_upstream,
```
renatav: I don't think commits like this are necessary if we describe what branches are in clone_from_disk

taf/git.py
```diff
@@ -191,26 +197,51 @@ def is_detached_head(self) -> bool:
 
     @property
     def is_git_repository(self) -> bool:
-        """Check if the given path is the root of a Git repository."""
+        """Check if the given path is inside a Git repository (work tree or bare)."""
         # This is used when instantiating a PyGitRepository repo, so do not use
-        # it here
+        # self.pygit / self.pygit_repo here (would recurse via the pygit property).
+        #
+        # Only a positive result is cached. A path that is not (yet) a repository
```
renatav: I find these comments too verbose, like above. E.g. Only a positive result is cached. A path that is not (yet) a repository can become one later enough

renatav (commented):
Some comments based a a partial review

taf/tuf/key_cache.py
```diff
@@ -0,0 +1,36 @@
...
+def _cached_crypto_key(self):
+    """Drop-in replacement for SSlibKey._crypto_key."""
+    return _load_pem_public_key_cached(self.keyval["public"])
+
+
+def enable_public_key_cache() -> None:
+    if hasattr(SSlibKey, "_crypto_key"):
+        SSlibKey._crypto_key = _cached_crypto_key  # type: ignore
+    else:
+        taf_logger.debug(
```
renatav: Should we maybe log this as info or warning? Though in that case we have to ensure it is only logged once during the run

taf/tests/test_git/test_git.py
```diff
@@ -663,3 +702,104 @@ def test_is_remote_branch(origin_repo: GitRepository, clone_repository: GitRepos
     assert not clone_repository.is_remote_branch(
         f"origin2/{clone_repository.default_branch}"
     )
+
+
+# --- pygit2-backed read-only checks ---
```
renatav: Nitpick: let's remove these comments

taf/tests/test_updater/test_handlers/test_in_memory_updater.py
```diff
@@ -0,0 +1,118 @@
...
+    assert fetcher.download_bytes("metadata/root.json", 100) == b"x" * 100
+
+
+def test_download_bytes_enforces_max_length():
+    fetcher = _FetcherStub(b"x" * 101)
+    with pytest.raises(DownloadLengthMismatchError):
+        fetcher.download_bytes("metadata/root.json", 100)
+
+
+@pytest.fixture
```
renatav: Let's move fixtures to `conftest` to stay consistent with the rest of the codebase

taf/tests/test_updater/test_update/test_update_partial.py
```diff
@@ -63,6 +64,42 @@ def test_update_partial_with_invalid_commits(origin_auth_repo, client_dir):
...
+    origin_auth_repo, client_dir
+):
+    clone_repositories(
+        origin_auth_repo,
+        client_dir,
+    )
+
+    client_auth_repo = AuthenticationRepository(path=client_dir / origin_auth_repo.name)
+    lvc_data = client_auth_repo.last_validated_data
+    lvc_data["exclude_filter"] = client_auth_repo.head_commit().value
```
renatav, line 86: Is there some execution path that is storing a commit as `exclude_filter`? That would indicate a different bug, even if not contained in `last_validated_data`

taf/tests/test_updater/test_update/test_update_valid.py
```diff
@@ -169,6 +170,44 @@ def test_update_valid_when_expiration_dates_updated(origin_auth_repo, client_dir
...
+
+    stale_origin = GitRepository(path=tmp_path / "stale-origin-auth")
+    stale_origin.clone_from_disk(
+        origin_auth_repo.path,
+        str(origin_auth_repo.path),
+        is_bare=True,
+    )
+
+    client_auth_repo = AuthenticationRepository(path=client_dir / origin_auth_repo.name)
+    client_auth_repo.set_remote_url(str(stale_origin.path))
```
renatav, line 198: If we can set this URL, but it is later ignored or overwritten, then we probably shouldn't even allow this?

renatav (commented):
A few more comments in case you are still around @n-dusan. Will continue a bit later

taf/updater/updater_pipeline.py
```diff
@@ -1302,9 +1334,18 @@ def set_excluded_targets(self):
                 last_validated_data = self.state.users_auth_repo.last_validated_data
                 # exclude filter can be specified if running local validation
                 if not self.exclude_filter:
-                    self.exclude_filter = (
-                        last_validated_data.get("exclude_filter") or None
-                    )
+                    stored_exclude_filter = last_validated_data.get("exclude_filter")
```
renatav, line 1337: This seems to be a workaround for a bug that appears elsewhere (maybe _reset_to_commits_before_pull, set_last_validated_of_repo). We should address the original bug. Since this probably isn't related to this PR, it does not have to be addressed here. I'd vote for removing this workaround and creating an issue for original problem.

renatav (commented):
One more comment. Made it through half of the PR.

taf/updater/updater_pipeline.py
```diff
@@ -2338,6 +2444,43 @@ def _get_repository_name_from_info_json(auth_repo, commit_sha):
...
+    speculative branch after a build), so the remote's HEAD is the
+    authoritative default branch (matches the clean-clone path). It is cached
+    and HEAD is repointed to it.
+    """
+    if users_auth_repo is None:
+        validation_auth_repo.clone(bare=True)
+        validation_auth_repo.fetch(fetch_all=True)
+        return
+
+    remote_url = urls[0] if urls else users_auth_repo.get_remote_url()
```
renatav, line 2469: Do we have a similar check appearing earlier in the pipeline? I would assume. If not, we should definitely move this to an earlier step

taf/git.py
```diff
@@ -191,26 +197,46 @@ def is_detached_head(self) -> bool:
...
+        # self.pygit / self.pygit_repo here (would recurse via the pygit property).
+        #
+        # Only a positive result is cached: a path that is not (yet) a
+        # repository can become one later.
+        if self._is_git_repository:
+            return True
+        # pygit2.discover_repository is a module-level function that searches
+        # upward for a repository, matching the semantics of the subprocess
+        # `rev-parse --is-inside-work-tree` / `--is-bare-repository` checks but
+        # without spawning a process (~150-400ms each on Windows).
```
renatav: Nitpick: we should remove the performance analysis comments from the code

taf/git.py
```diff
@@ -191,26 +197,46 @@ def is_detached_head(self) -> bool:
...
+        if PYGIT2_AVAILABLE and not self.allow_unsafe:
```
renatav: I think it's fine to start removing pygit2 fallbacks from the code. We now have a check that would raise an error if it is not installed. No one has had any issues with pygit2 for a very long time I think.

taf/git.py
```diff
@@ -317,7 +343,16 @@ def _get_default_branch_from_local(self) -> str:
         In those cases, get the default branch from `git remote show origin`.
 
         If the repository does not have a remote set, use `symbolic-ref` HEAD
+
+        The cheap symbolic-ref reads (steps 1 and 3) are done in-process via
+        pygit2 when possible to avoid spawning subprocesses (~150-400ms each on
```
renatav: Nitpick: remove performance analysis comments

taf/git.py
```diff
@@ -1787,10 +1958,22 @@ def something_to_commit(self) -> bool:
         if self.is_bare_repository:
             # For bare repositories, use `git diff` to check for uncommitted changes
             uncommitted_changes = self._git("diff --name-only --cached")
-        else:
-            # For non-bare repositories, use `git status`
-            uncommitted_changes = self._git("status --porcelain")
-        return bool(uncommitted_changes)
+            return bool(uncommitted_changes)
+        # For non-bare repositories, check status in-process via pygit2 first.
```
renatav, line 1960: This would be cleaner without the `pygit2` availability check, which I think we can slowly start removing from the code.

taf/git.py
```diff
@@ -1881,16 +2070,46 @@ def _determine_default_branch(self) -> Optional[str]:
...
+        # Do not cache None — the repo may not exist yet (e.g. temp path at
+        # instantiation time). The next call after the repo is created will
+        # detect the branch correctly instead of returning a stale None.
         return None
 
+    def clone_bare_from_local(self, local_path: Path) -> None:
+        """Seed a bare repo from a local path (no network).
+
+        Uses `git clone --local`, which hardlinks objects on the same volume
+        instead of copying them. Does NOT detect default_branch — the caller
```
renatav: Nitpick: I personally do not like when LLMS capitalize NOT or any other word, it looks aggressive imo, so I'd personally be mindful of that

taf/git.py
```diff
@@ -1881,16 +2070,46 @@ def _determine_default_branch(self) -> Optional[str]:
...
+            str(local_path),
+            error_if_not_exists=False,
+            reraise_error=True,
+        )
+        self._is_git_repository = None
+        # `git clone --bare` creates no origin remote (pygit2 did); add it for
+        # parity so the caller's `remote set-url origin` succeeds
+        self.add_remote("origin", str(local_path))
+        # Clear cache and reset so the caller's post-setup detection runs cleanly.
+        _default_branch_cache.pop(str(self.path), None)
```
renatav: Nitpick: maybe have a clear default branch helper that would always perform both of these operations

renatav (commented):
Final comments

taf/tests/test_updater/test_update/test_update_partial.py
n-dusan: The cause is `_reset_to_commits_before_pull`, which calls a method (`set_last_validated_commit`) that was removed in an earlier refactor, so it errors instead of corrupting data. Only runs in dev mode after a handler failure. So appears to be pre-existing and not from this PR. I'd keep the small guard and test and open a separate issue?

taf/tests/test_updater/test_update/test_update_valid.py
n-dusan: The stored URL isn't ignored. It's the fallback when no remote is passed. The test sets a stale one and passes a real one to show the passed-in remote is taken first

taf/updater/updater_pipeline.py
n-dusan: Dropped it. The URL is already validated in `clone_repository`/`update_repository` before the pipeline runs, so the check here was unnecessary

renatav (approved):
Created an issue for the workaround added in this PR, as the original issue should be addressed and the workaround should be removed. Tested Ubuntu.

PR #748, opened 2026-06-19
Dgreisen llm/fixes

dgreisen-llm:

various ux fixes

### DIFF
--- a/taf/config.py
+++ b/taf/config.py
@@ -9,24 +9,39 @@ from taf.config import load_config
 cfg = load_config()
 cfg = load_config("path/to/config.toml")
 
-print(cfg.root.org)                     # -> "some-repo-law-org"
+print(cfg.root.name)                    # -> "some-repo-law"
 """
 
 from __future__ import annotations
 
 import attrs
+import cattrs
 import tomli as toml  # type: ignore[import-untyped]
 from pathlib import Path
 from typing import Any, Mapping, Optional
 
+from taf.exceptions import InvalidConfigError
+
 
 @attrs.define(slots=True, frozen=True, kw_only=True)
 class RootConfig:
-    """Representation of the TOML `[root]` table."""
+    """Representation of the TOML `[root]` table.
+
+    The `[root]` table is optional: a root auth repo carries significant trust,
+    so declaring one must never be a precondition for unrelated archive work.
+    When the table *is* present it identifies the root authentication repo, so
+    `name` and `org` are required.
+
+    `url` and `hash` are optional records: the archive tooling writes `url`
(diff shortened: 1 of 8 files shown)


taf/tests/tuf/test_keys/test_yubikey_setup.py
```diff
@@ -0,0 +1,81 @@
...
+
+import pytest
+
+import taf.yubikey.yubikey as yk
+from taf.yubikey.yubikey import yubikey_prompt
+from taf.yubikey.yubikey_manager import PinManager
+
+
+@pytest.fixture
+def fake_single_yubikey(monkeypatch):
```
renatav: I think that we should be able to reuse `FakeYubiKey`, but extend it to make key loading optional. Alternatively, we could move `fake_single_yubikey` to `conftest` for future use.

taf/config.py
```diff
@@ -9,24 +9,40 @@
...
-    """Representation of the TOML `[root]` table."""
+    """Representation of the TOML `[root]` table.
+
+    The `[root]` table is optional: a root auth repo carries significant trust,
+    so declaring one must never be a precondition for unrelated archive work.
+    When the table *is* present it identifies the root authentication repo, so
+    `name` and `org` are required.
+
+    `url` and `hash` are optional records: the archive tooling writes `url`
+    (the root stele clone URL) but nothing reads it back, and `hash` is not
```
renatav: Nitpick: In TAF, we don't use the "stele" terminology, so this docstring and the following one might be confusing to people.

renatav (commented):
Makes sense to me. Could you also update the changelog, please?

dgreisen-llm:
@renatav thanks for the review — addressed all three points:

- **Changelog**: added entries under `[Unreleased] → Fixed` for the three fixes in this PR ([commit](https://github.com/openlawlibrary/taf/pull/748/commits)).
- **`config.py` "stele" terminology**: dropped `stele`/`stelae` in favor of TAF-neutral phrasing ("the root repository's clone URL", "the archive tooling").
- **`fake_single_yubikey` fixture**: moved into `taf/tests/tuf/test_keys/conftest.py` for reuse. I kept the `taf.yubikey.yubikey` import lazy (inside the fixture) so the rest of the directory's tests still collect without the optional `ykman` extra installed. I went with relocating the fixture rather than extending `FakeYubiKey`, since this test only needs `get_serial_nums` patched and `FakeYubiKey`'s key-loading interface would be overkill here — happy to switch to extending it if you'd prefer.

Ready for another look when you have a moment.

renatav (approved):

Issue #698, opened 2026-01-26
`taf roles list-keys`: Add role name to output when listing all keys


n-dusan:
Currently `taf roles list-keys` without a role argument lists all key IDs but doesn't show which role each key belongs to, making the output less useful than listing a specific role.

Expected behavior:
```
Role: docs
Key ID: ba65808977548012b65ab8229f7b939e2339703cec3686ef816f2480274e7f9b

Role: targets
Key ID: a1b2c3d4...
```

Current behavior:

Only key IDs are printed with no role association.

PR #755, opened 2026-07-03
Issue #698: Fix roles list keys

ZCute1:
## Description

Fix `taf roles list-keys` when called without a role argument so it prints each key with its associated role(s), instead of only showing key IDs.

Also added CLI test coverage for `taf roles list-keys --path <repo>` with no role positional argument.

### DIFF
--- a/taf/api/roles.py
+++ b/taf/api/roles.py
@@ -917,38 +917,68 @@ def _get_roles_key_size(role: str, keystore: str, keys_num: int) -> int:
     return get_key_size(key_path)
 
 
+def _format_role_keys_block(role_label: str, key_ids: List[str], certs_dir: str) -> str:
+    key_lines = [
+        str(get_metadata_key_info(certs_dir, key_id)) for key_id in key_ids
+    ]
+    return f"Role: {role_label}\n" + "\n".join(key_lines)
+
+
+def _roles_for_keys(auth_repo: AuthenticationRepository) -> Dict[str, set]:
+    roles_for_key: Dict[str, set] = defaultdict(set)
+    for role_name in auth_repo.get_all_roles():
+        role_key_ids = auth_repo.get_role_keys(role=role_name)
+        if role_key_ids:
+            for key_id in role_key_ids:
+                roles_for_key[key_id].add(role_name)
+    return roles_for_key
+
+
 @log_on_error(
     ERROR,
-    "Could not list keys of {role:s}: {e}",
+    "Could not list keys: {e}",
     logger=taf_logger,
     on_exceptions=TAFError,
     reraise=True,
 )
 def list_keys_of_role(
     path: str,
-    role: str,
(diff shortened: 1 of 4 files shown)


taf/api/roles.py
```diff
@@ -917,38 +917,68 @@ def _get_roles_key_size(role: str, keystore: str, keys_num: int) -> int:
...
 
 
+def _format_role_keys_block(role_label: str, key_ids: List[str], certs_dir: str) -> str:
+    key_lines = [
+        str(get_metadata_key_info(certs_dir, key_id)) for key_id in key_ids
+    ]
+    return f"Role: {role_label}\n" + "\n".join(key_lines)
+
+
+def _roles_for_keys(auth_repo: AuthenticationRepository) -> Dict[str, set]:
```
renatav, line 927: It was surprising that a similar utility was not already somewhere in the codebase. Let's move this to `tuf/repository/MetadataRepository`

taf/api/roles.py
```diff
@@ -917,38 +917,68 @@ def _get_roles_key_size(role: str, keystore: str, keys_num: int) -> int:
...
+    roles_for_key: Dict[str, set] = defaultdict(set)
+    for role_name in auth_repo.get_all_roles():
```
renatav, line 929: This works, but is not the most efficient solution. That is fine as we do not expect to ever encounter a situation where we have hundreds of roles. 

`get_all_roles()` walks the full delegation tree and in doing so, it already knows each role's parent, but throws that away and returns a flat list. Then, for each of those role names, `get_role_keys(role_name)` has to call `find_delegated_roles_parent(role)`. Implementing something similar to `find_keysid_roles` would be more efficient.

However, this is fine as is, unless you want to tackle the harder implementation as a way of getting more familiar with the system.

taf/api/roles.py
```diff
@@ -917,38 +917,68 @@ def _get_roles_key_size(role: str, keystore: str, keys_num: int) -> int:
...
+
 @log_on_error(
     ERROR,
-    "Could not list keys of {role:s}: {e}",
+    "Could not list keys: {e}",
     logger=taf_logger,
     on_exceptions=TAFError,
     reraise=True,
 )
 def list_keys_of_role(
```
renatav, line 944: I think it's actually better to split the two up:kKeep list_keys_of_role(path, role) exactly as it was (role required), and add a new function for the "all roles" case `list_keys_by_roles`. This preserves the existing function's contract, and the existing tests should pass without needing any modifications. And then the CLI could call one or the other.

renatav (commented):
I think is very close to being what we need. My main concern is keeping the public api as intuitive and clean as possible, so it might be a good idea to refactor this a bit before merging.

renatav (approved):
I'm going to merge this in as it is an improvements over how the command works now.

PR #763, opened 2026-08-06
fix: make rotating file logs tolerate the log file being held

n-dusan:

taf can end up sharing its log files across processes - e.g. a long-running host process that imports taf directly, alongside short-lived `taf` CLI subprocesses writing to the same ~/.taf log files. On Windows, if one process still has the file open, another process's rotation attempt fails with a PermissionError on rename. loguru's built-in rotation isn't resilient to that: the failed rename leaves the sink's file handle closed, so every subsequent write retries the same doomed rotation and fails again, flooding stderr with "Logging error" tracebacks for the rest of the process's life.

Replace the plain-path sink with a small rotating sink that treats a failed rotation as non-fatal - it keeps appending to the existing file and only retries rotation after a cooldown, instead of on every message.

### DIFF
--- a/taf/log.py
+++ b/taf/log.py
@@ -1,7 +1,10 @@
 import os
 import sys
+import time
 import logging
-from typing import Dict
+import zipfile
+import datetime
+from typing import Dict, Optional, TextIO, Union
 from pathlib import Path
 
 from loguru import logger as taf_logger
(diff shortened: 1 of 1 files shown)


n-dusan:
Closing in favor of #764

PR #766, opened 2026-08-10
Support Git LFS in target repositories

dgreisen-llm:
## Description

Git LFS was entirely uncovered here — no test, no issue, no PR mentioned it — so whether TAF worked with
LFS target repositories was unknown. It turns out it does over the wire, with one constraint worth
documenting, and pygit2 checkouts did not, which this fixes.

### What changed

1. **`test:` deterministic git environment.** 20 `test_git` tests failed on any machine setting
   `init.defaultBranch`. Git's repository discovery walks *upward*, and test repositories live inside the
   TAF checkout, so detection on a not-yet-created repo answered about **TAF's own repository**. CI passed
   only because the enclosing branch happened to match the `git init` default — on a branch named anything
   else, those 20 would have failed there too. Fixed with `GIT_CEILING_DIRECTORIES`, `GIT_CONFIG_GLOBAL`
   and `GIT_CONFIG_NOSYSTEM` in a session fixture.
2. **`build:` one `pygit2==1.14.*` pin.** The `< 3.11 → 1.9.*` split dates from #440 and was never chosen
   on the merits; 1.14 ships cp39–cp312 wheels and covers the whole tested matrix. `python_requires`
   and the classifiers now match CI.
3. **`test:` LFS coverage**, including an in-process Git LFS server speaking the real Batch API, so no
   service has to be installed to run the suite. CI gains `git-lfs` from apt.
4. **`feat:` `taf.lfs`**, a libgit2 filter that pipes LFS blobs through the `git-lfs` binary.
5. **`fix:` cap `securesystemslib` below 1.5** — unrelated to LFS, but CI cannot pass without it. See
   "The dependency break" below.

### The LFS picture

TAF never clones a target repository straight from its origin: it stages through a **bare** intermediate
clone, which carries no LFS objects. So with an LFS server configured (via committed `.lfsconfig` — the
deployed shape) clone and update work, proven by deleting the origin's local objects and asserting the
server served them. With no `lfs.url` the update fails; that is a documented constraint, now in the
CHANGELOG.

`taf.lfs` covers the other half: libgit2 does not run the external filters declared in git config, so
pygit2 checkouts wrote pointer text. The filter runs only for paths `.gitattributes` routes to LFS, and
only in the direction git itself filters in that repository.

It talks to **one long-running `git-lfs filter-process`** per operation over git's own pkt-line protocol,
as git does, and stages content in memory up to 5 MB and on disk beyond. That is what keeps an archive of
100k small files from paying a process spawn each, and a 200 MB PDF from being held in RAM. Both are
measured by tests, not asserted in prose:

| performance test | budget it enforces |
| --- | --- |
| `test_a_large_file_is_not_held_in_memory` | 200 MB file checked out under 64 MB peak RSS (measured in a subprocess, `VmHWM` reset first) |
| `test_many_files_share_one_git_lfs_process` | 41 LFS files ⇒ at most 2 `git-lfs` spawns, counted by a shim on `PATH` |

### The dependency break

CI on this branch first failed with 19 failures and 225 errors that had nothing to do with it. Pushing
**unmodified master** as a baseline reproduced them exactly, so they were master's, not the branch's:
`securesystemslib` 1.5.0 requires `cryptography>=48`, `setup.py` pins `cryptography==43.0.*`, and
`pyOpenSSL==24.2.*` caps it below 44. The bound lives on an extra pip is not asked to resolve, so nothing
fails at install time — it surfaces at runtime as `UnsupportedLibraryError: 'pyca/cryptography' library
required`, which in the suite turns into `pytest: reading from stdin while output is captured!` from a
`click.confirm` in `taf/keystore.py`. `securesystemslib>=1.3,<1.5` restores it; 1.4.0 verified working.

## Review history

Repeated rounds of adversarial review with fresh reviewers, each finding verified by reproduction and
by mutation testing in a scratch worktree. Later rounds found defects in earlier rounds' fixes.

Defects found and fixed: an `ImportError` when pygit2 is absent; `init_repo` turned into a network call
that could hang; **0-byte truncation of working-tree files** when a smudge failed; an **uncommitted edit
destroyed** when a clean failed; the filter running where git does not; `--skip-smudge` disabling clean
and so committing raw content where a pointer belongs; legacy `filter.lfs.smudge`/`clean` config ignored,
silently checking out pointer text; running git-lfs for a `filter.lfs.*` that names some other program; a
filter-process pool that lived as long as the interpreter, leaking one git-lfs per repository and raising
a **false error on a successful checkout** from a stale failure record; and a config cache keyed on
guessed paths that missed `[include] path` and linked worktrees.

Reviewers also caught defects in the *tests*: a clean-failure test that passed whether or not the clean
failed, a memory measurement inflated by `ru_maxrss` inheritance across fork and exec, a harness that read
a 200 MB file into a `repr()`, and a helper that simulated an absent git-lfs by dropping every PATH
directory that provides it — which on CI drops `git` too, since apt installs both into `/usr/bin`.

The last round found the most serious defect of all, and it is worth stating plainly: **libgit2 also runs
the clean filter to answer `repo.status()`**, which `something_to_commit` called with no session around
it. A failure recorded there had nobody to collect it, so the *next* checkout of that repository raised
it — blaming the LFS server for files that checkout had just materialized correctly. The updater reaches
this: `updater_pipeline` asks every target repository for uncommitted changes before it checks anything
out. The same gap cost a process per file: **42 git-lfs launches for 41 modified files, now 2**. A failure
record now lives no longer than the session that made it, and the status and merge paths join a session
like the four checkout paths.

The round after it found a regression in that fix, which is the honest shape of this work: wrapping
`repo.merge` in a session put the failure *between* the merge and its commit, so `GitLFSError` left
`MERGE_HEAD` and a staged index behind and every later merge in that repository failed. Committing and
cleaning up inside the session fixes it. The same round found that the headline fix was not pinned — the
test passed when any single mechanism was reverted, only failing when both were — and that a
`raise_on_failure` flag I had added decided nothing, because its one caller already wrapped the call in
`except Exception: pass`. Five mutations that had left the whole suite green now each fail one test.

That earlier round also removed the retry-once in `GitLFSProcess.filter`. It fired on the common path, not on a
stale connection — libgit2's clean pass warms the process, so the first smudge counts as "reused" — and
git-lfs exits rather than reporting a per-file error, so the retry asked the Batch API twice for an object
it had already refused. Removing it also removed the only route to a silent truncation. Two more tests
turned out to be unfalsifiable: the spawn-count test passed with the filter entirely disabled, and
`is_lfs_pointer` called empty content a pointer, because `git lfs pointer --check` exits 0 on empty input,
which left the tolerated-failure test blind to a zero-byte file.

## Three things for the maintainer

**1. `PYGIT2_AVAILABLE` — done, not asked.** You told us in review that pygit2 should not be treated as an
optional dependency, and I had been carrying an `if PYGIT2_AVAILABLE:` guard around the LFS import anyway,
justified by an `ImportError` I produced myself with a meta-path blocker. `setup.py` requires pygit2
unconditionally and no test needed the guard, so it is gone: `taf/git.py` imports `filtering` directly.
One consequence to note — the `try: import pygit2 / except ImportError` block above it is now unreachable
dead code, along with the three `if not PYGIT2_AVAILABLE:` fallbacks it feeds. Removing those means
rewriting two tests that flip the flag to exercise the subprocess paths, so I left it; say if you want it
in this PR.

**2. What the filter still cannot do.** Four limits, all verified rather than assumed:

- **No delayed batching.** libgit2's filter API is synchronous per blob, so git's `capability=delay` —
  which lets git-lfs fetch many objects in one Batch API round trip — is unavailable. Objects already in
  the local store cost nothing extra; a cold checkout makes one round trip per object.
- **A filter cannot abort the operation.** Raising inside one reaches libgit2 as `failed to close filter
  stream`, carrying none of the reason, and whatever was already forwarded stays in the destination file.
  (Measured on pygit2 1.14.1 / libgit2 1.7.2: a filter that forwarded 7 bytes and then raised left a
  1400-byte file at 7 bytes.) So nothing is forwarded until the whole result is in hand, the unfiltered
  blob is written through, and the caller raises once the operation finishes. Working-tree files are
  always readable — a pointer, or the bytes that were already there.
- **No fail-fast.** With the server unreachable, every remaining file is attempted; git aborts at the
  first failure. Telling "this object is missing" from "the server is gone" is not reliable, and passing
  everything through after one miss would leave pointers for files that were fetchable, so this is left as
  it is. The reporting is bounded instead: one error line per operation with the failed count, not one per
  file.
- **Temporary space.** Streaming keeps a 200 MB object out of RAM but stages it on disk first, under
  `TMPDIR`, in full. Now in the CHANGELOG. Staging inside `.git` instead would remove the requirement
  altogether; it is a separate change, and I could not make the current failure mode testable (an
  unwritable `TMPDIR` is not reachable — Python falls back to `/tmp`), so I did not add an untested error
  path for it.

**3. History.** 23 commits, including one experiment and its revert (`7afc8fed` / `de5e0724`), kept so the
review trail is legible. Say the word and I'll squash. Three commit messages are inaccurate and are too
far back to rewrite: `e7cc1cb7` describes about a quarter of its diff, `2f7c8635` says "twenty tests" where
the count is 17, and `72e1fd1f` claims its comments state what the code does while introducing four that
narrate the change it made. `bbc130aa`'s subject — dropping the `PYGIT2_AVAILABLE` guard — is also no
longer true of the final tree, since the guard came back in `08616a4b` for the reason in item 1. A fourth,
one commit old, said "two performance tests" where it was one; that one I amended rather than leave
standing. Every one of them was found by a reviewer, not by me.

## An intermittent CI failure, unexplained

One 3.10 job failed once on `test_update_with_invalid_dependency_repo`
([run 32742201079](https://github.com/dgreisen-llm/taf/actions/runs/32742201079)) with `DID NOT RAISE` —
the updater *accepted* a dependency whose timestamp signature had been deliberately corrupted. Rerunning
the same commit passed all three jobs, and it does not reproduce locally: 40 single-test runs on 3.10, 15
runs of its module, a full 3.10 suite and two full 3.12 suites, all green.

I could not attribute it. The fixture commits its corruption only if `something_to_commit()` reports dirty
(`taf/api/api_workflow.py:319`), and that method trusts libgit2's *clean* verdict, so a wrong "clean"
would produce exactly this. I tested that: 300 trials of a same-size rewrite inside the index's own second
— libgit2's `status()` missed the change zero times. Hypothesis refuted.

Evidence against this branch being involved: the only overlap is the `filtering()` wrapper now around that
status call, and for a repository with no `filter=lfs` attribute the filter's `check()` is never invoked,
so nothing is read or spawned; a reviewer separately confirmed `something_to_commit` returns identically to
master across six LFS/non-LFS clean/dirty combinations. That is evidence, not proof — I am flagging it
rather than treating a green rerun as an answer, because a validation test that intermittently fails to
notice tampered metadata is the failure mode this project exists to prevent.

## Two pre-existing bugs found in `merge_branch`

Both are identical on master, unrelated to Git LFS, and left alone here:

- **A conflicting merge silently commits the conflict markers.** libgit2's `merge` returns 0 with conflicts
  staged, then `commit()`'s `git add -A` stages the marker text and the commit succeeds — where plain git
  refuses with "Committing is not possible because you have unmerged files". `merge_branch` reports
  success and `HEAD` holds `<<<<<<<`.
- **An unknown branch name raises `AttributeError: 'NoneType' object has no attribute 'target'`** —
  `lookup_branch` returns None and `.target` is taken unguarded. The same "Cannot None" class of message
  `test_clone_errors.py` exists to prevent.

## Testing

| run | result |
| --- | --- |
| CI, Python 3.10 / 3.11 / 3.12 | green (see the intermittent failure above) |
| full suite, Python 3.10 and 3.12 | 439 passed, 3 skipped |
| LFS suite (`test_lfs*`, `test_lfs_protocol`) | 52 tests |
| LFS suite, git-lfs absent | 14 passed, 28 skipped — the pkt-line and config tests need no binary |
| mutation testing, scratch worktree | every mutation tried killed a specific test; a few narrow error paths (a mid-stream per-file error, the handshake greeting) are covered only indirectly |
| black / flake8 / bandit / mypy | clean, under Python 3.10, 3.11 and 3.12 |

flake8 5.0.4 rejects a semicolon inside an f-string on 3.12 only (PEP 701 changed how f-strings tokenize),
which is why hook cleanliness is claimed per interpreter.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SFYACn5YKRCCqtUedS8aFj


taf/tests/test_updater/test_lfs/lfs_server.py
```diff
@@ -0,0 +1,213 @@
...
+import shutil
+import threading
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from pathlib import Path
+from typing import Dict, List, Optional
+
+LFS_CONTENT_TYPE = "application/vnd.git-lfs+json"
+
+
+class _LFSRequestHandler(BaseHTTPRequestHandler):
```
renatav, line 29: I think we should try to get the LLMs to follow our existing convention of ordering methods/functions alphabetically. I’ve always thought that some exceptions to that rule were warranted, but LLM's ordering often seem random

taf/tests/test_updater/lfs_utils.py
```diff
@@ -0,0 +1,205 @@
...
+        ["git"] + args,
+        cwd=str(cwd),
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return result.stdout
+
+
+def add_lfs_target_commits(
```
renatav: `add_lfs_target_commits` / `update_lfs_target_commits` feel a bit too parallel to the existing target-repo helpers. We already have shared test helpers for “mutate target repos, commit, then sign”, but those are currently hard-coded around generating/appending random text. Rather than adding a separate LFS-only path here, I think it would be cleaner to generalize the existing helpers so they can accept an explicit mutation or file content, and then reuse that flow from the LFS tests. T

taf/tests/test_updater/lfs_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    from. This is how hosted LFS repositories are normally set up, and TAF
+    depends on it (see ``test_lfs.py``).
+    """
+    _run_git(repo.path, ["lfs", "install", "--local"])
+    _run_git(repo.path, ["lfs", "track", LFS_TRACKED_PATTERN])
+    if lfs_url is not None:
+        (repo.path / ".lfsconfig").write_text(f"[lfs]\n\turl = {lfs_url}\n")
+
+
+def _run_git(cwd: Path, args: List[str]) -> str:
```
renatav: We already have shared command helpers in `taf.utils`, so I'd prefer to reuse them, extend them if necessary or move a helper like this there.

taf/tests/test_updater/lfs_utils.py
```diff
@@ -0,0 +1,205 @@
...
+from taf.auth_repo import AuthenticationRepository
+from taf.git import GitRepository
+from taf.tests.conftest import KEYSTORE_PATH, TEST_DATA_ORIGIN_PATH
+from taf.tests.test_updater.conftest import sign_target_repositories
+from taf.yubikey.yubikey_manager import PinManager
+
+LFS_POINTER_PREFIX = b"version https://git-lfs.github.com/spec/v1"
+
+#: Pattern committed to ``.gitattributes`` so git routes these files to LFS.
+LFS_TRACKED_PATTERN = "*.bin"
```
renatav: These constants feel more like test scenario data than utilities. LFS_TRACKED_PATTERN, LFS_FILE_NAME, and PLAIN_FILE_NAME define the shape of the LFS test setup, so I’d expect them to live alongside the tests (or in a conftest) rather than in a utils module

taf/tests/test_updater/test_lfs.py
```diff
@@ -0,0 +1,368 @@
...
+    with ThreadPoolExecutor(max_workers=4) as executor:
+        list(executor.map(checkout, repos))
+
+    assert not errors, f"concurrent checkouts failed: {errors}"
+    for repo in repos:
+        assert_lfs_content_materialized(repo.path, "v2")
+
+
+@pytest.mark.parametrize("origin_auth_repo", [TARGETS_WITH_LFS], indirect=True)
+def test_cloned_lfs_file_is_tracked_by_lfs_at_all(origin_auth_repo, client_dir):
```
renatav: If this is guarding the validity of the whole LFS test setup, I’m not sure it wants to be a standalone test at the end. Could this assertion live in the shared LFS setup/helper instead, so every test gets that guarantee directly?

taf/tests/conftest.py
```diff
@@ -29,9 +29,57 @@
...
+    monkeypatch.setenv("GIT_CONFIG_GLOBAL", str(config_path))
+    monkeypatch.setenv("GIT_CEILING_DIRECTORIES", str(TEST_DATA_PATH.parent))
+    yield config_path
+    monkeypatch.undo()
+    shutil.rmtree(config_dir, ignore_errors=True)
+
 
 @pytest.fixture(scope="session", autouse=True)
-def repo_dir():
+def repo_dir(deterministic_git_environment):
```
renatav, line 98: I think this is actually a really good addition and now that I see this, I am asking myself why none of us came up with it earlier.
Nitpick: I would personally prefer something like this:
```
@pytest.fixture(scope="session", autouse=True)
def repo_dir():
    with deterministic_git_environment_context():
        ...
        yield path
```

taf/git.py
```diff
@@ -44,6 +45,14 @@
     pygit2 = None
     PYGIT2_AVAILABLE = False
 
+# Registered here, at import time, because pygit2's filter registry is not
+# thread-safe and the updater materializes target repositories concurrently -
+# so this has to happen before any repository object or worker thread exists.
+# Without it, pygit2 checkouts write Git LFS pointer text into the working tree
+# instead of file content. Called unconditionally: `taf.lfs` reports whether it
+# could register (it needs pygit2 >= 1.13.3) and is a no-op otherwise.
+register_lfs_filter()
```
renatav: Would it be possible for `taf.lfs` to register the filter on import?

taf/lfs.py
```diff
@@ -0,0 +1,182 @@
...
+import shutil
+import subprocess
+from functools import lru_cache
+from typing import Optional
+
+from taf.exceptions import GitLFSError
+from taf.log import taf_logger
+
+try:
+    import pygit2
```
renatav: I don't think there is a need to treat `pygit2` as an optional dependency. We've actually been moving away from that as we haven't had any issues with `pygit2` in over 5 years probably.

taf/lfs.py
```diff
@@ -0,0 +1,182 @@
...
+POINTER_PREFIX = b"version https://git-lfs"
+
+_registered = False
+
+
+@lru_cache(maxsize=1)
+def git_lfs_executable() -> Optional[str]:
+    """Path to the ``git-lfs`` binary, or None when it is not installed.
+
+    Cached because it is consulted once per filtered file. Tests replace this
```
renatav: The last sentence, explaining what tests are doing doesn't belong here imo. I've noticed that LLMs often include inappropriate explanations (e.g. a correction I made, something narrowly related to a bug that was recently fixed)

taf/lfs.py
```diff
@@ -0,0 +1,182 @@
...
+        def write(self, data, src, write_next):
+            self._chunks.append(bytes(data))
+
+        def close(self, write_next):
+            payload = b"".join(self._chunks)
+            if not payload:
+                return
+
+            verb = "smudge" if self._mode == pygit2.GIT_FILTER_SMUDGE else "clean"
+            executable = git_lfs_executable()
```
renatav: Something like `get_git_lfs_executable` would be more in line with our naming conventions

taf/lfs.py
```diff
@@ -0,0 +1,182 @@
...
+                    "Cannot check out {}: it is tracked by Git LFS but the "
+                    "git-lfs binary was not found on PATH. Install Git LFS "
+                    "(https://git-lfs.com) and retry.",
+                    self._path,
+                )
+                raise GitLFSError(
+                    f"git-lfs is required to check out {self._path} but is not installed"
+                )
+
+            try:
```
renatav: This works, but I’d prefer not to have the raw subprocess.run(...) embedded here. Could we extract the git-lfs invocation into a helper so the main filter logic reads at a slightly higher level?

taf/lfs.py
```diff
@@ -0,0 +1,182 @@
...
+                stderr = result.stderr.decode(errors="replace").strip()
+                taf_logger.error(
+                    "git-lfs {} failed for {}: {}", verb, self._path, stderr
+                )
+                raise GitLFSError(f"git-lfs {verb} failed for {self._path}: {stderr}")
+
+            if verb == "smudge" and result.stdout.startswith(POINTER_PREFIX):
+                # git-lfs exits 0 and echoes the pointer back when it cannot
+                # obtain the object. Refuse it rather than write it to disk.
+                taf_logger.error(
```
renatav: This looks like it could be user-visible, so I don’t think the wording should be this technical/internal. Same with other error messages in this module

setup.py
```diff
@@ -63,8 +63,10 @@
         "cryptography==43.0.*",
         "securesystemslib==1.*",
         "loguru==0.7.*",
-        'pygit2==1.9.*; python_version < "3.11"',
-        'pygit2==1.14.*; python_version >= "3.11"',
+        # One pin for every supported Python: 1.14 ships wheels for cp39-cp312,
```
renatav: This comment is PR-explanation/commentary that LLMs like to add, not really source-code help.

setup.py
```diff
@@ -92,7 +94,6 @@
         "License :: OSI Approved :: GNU Affero General Public License v3 or later (AGPLv3+)",
         "Operating System :: OS Independent",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
```
renatav: Let's remove 3.9 too, we no longer support it either.

taf/lfs.py
```diff
@@ -0,0 +1,182 @@
...
+                raise GitLFSError(f"could not run git-lfs {verb}: {error}") from error
+
+            if result.returncode != 0:
+                stderr = result.stderr.decode(errors="replace").strip()
+                taf_logger.error(
+                    "git-lfs {} failed for {}: {}", verb, self._path, stderr
+                )
+                raise GitLFSError(f"git-lfs {verb} failed for {self._path}: {stderr}")
+
+            if verb == "smudge" and result.stdout.startswith(POINTER_PREFIX):
```
renatav: This `POINTER_PREFIX` seems really broad, are we sure this is not going to detect legitimate file content as a failed fetch?

renatav (commented):
This makes sense to me overall, and keeping it fairly separate from the older code paths makes me much less nervous about merging it. I’m leaving a few comments, mostly around cleanup, consistency, and some generated code style patterns that I think are worth tightening before merge.

taf/tests/test_updater/test_lfs/test_lfs.py
```diff
@@ -0,0 +1,842 @@
...
+    A merge is only over once it is committed and its state cleaned up. Failing
+    before that leaves ``MERGE_HEAD`` behind, and every later merge in that
+    repository fails.
+    """
+    origin = build_lfs_origin(tmp_path / "origin")
+    client = GitRepository(path=tmp_path / "client")
+    client.clone_from_disk(origin.path, keep_remote=True)
+    client.checkout_branch("other")
+
+    run("git", "-C", str(client.path), "config", "lfs.url", "http://127.0.0.1:1/lfs")
```
renatav, line 117: Can we reuse GitRepository's `_git`?

taf/exceptions.py
```diff
@@ -56,6 +56,10 @@ def __init__(
         self.error = error
 
 
+class GitLFSError(TAFError):
```
renatav, line 59: Every git error already at the moment inherits from GitError, so GitLFSError(TAFError) breaks that pattern.  It also seems like  GitLFSError failures aren't being handled consistently, and some paths can end up swallowing or missing them. Might be worth another pass

taf/lfs.py
```diff
@@ -0,0 +1,370 @@
+"""Make libgit2 checkouts honor Git LFS, by delegating to the git-lfs binary.
```
renatav, line 1: Since lfs.py, lfs_process.py, and lfs_protocol.py are three tightly-coupled files, would it make sense to group them into a taf/lfs/ package, same as taf/yubikey/ already does with yubikey.py/yubikey_manager.py?

renatav (commented):
I haven't done a deep, line-by-line review of the new LFS process/protocol code itself as that would need more time to really get familiar with it. I did review its impact on the existing codebase and checked it against our conventions.

Given the LFS internals are new and maybe someone else is currently working in this area should also take a pass before this merges

taf/tests/test_updater/test_lfs/lfs_server.py
dgreisen-llm: Done — functions, methods and classes are alphabetical across the LFS modules now (dunders first, then public, then private). Verified with an AST check rather than by eye, because I broke it once with a rename and didn't notice.

taf/tests/test_updater/lfs_utils.py
dgreisen-llm: Done — the LFS-only helpers are gone. `add_valid_target_commits` / `update_target_repository` now take an optional `mutate` callable plus `commit_message`, and the LFS tests pass their own mutation into them.

taf/tests/test_updater/lfs_utils.py
dgreisen-llm: Partly, and honestly. `lfs_utils.py` and its duplicate git runner are gone. The one direct `subprocess.run` left is the `git config` read inside the filter: `taf.utils.run` appends `--progress` when given a timeout, and a timeout matters there since it runs inside a checkout. Your 8/25 note about reusing `_git` in the tests is next on my list and I'll revisit this at the same time.

taf/tests/test_updater/lfs_utils.py
dgreisen-llm: Done — `LFS_TRACKED_PATTERN`, `LFS_FILE_NAME` and `PLAIN_FILE_NAME` live in the LFS tests' conftest, and the targets config sits in the test module that uses it.

taf/tests/test_updater/test_lfs.py
dgreisen-llm: Done — the assertion moved into `build_lfs_origin`, so every test that builds an LFS origin gets the guarantee instead of one test at the end.

taf/tests/conftest.py
dgreisen-llm: Done, close to your shape: `deterministic_git_environment_context()` is the context manager, wrapped by a session fixture that `repo_dir` depends on rather than folded into `repo_dir` itself. Say if you'd prefer it inline there.

taf/git.py
dgreisen-llm: Yes — `taf.lfs` calls `pygit2.filter_register` at import, and `taf/git.py` only imports `filtering`.

taf/lfs.py
dgreisen-llm: Done, belatedly — this one is on me. `taf.lfs` imports pygit2 unconditionally, and as of e3d6fc0 so does `taf/git.py`; the `if PYGIT2_AVAILABLE:` guard I had kept around the LFS import is gone. Consequence worth naming: the `try: import pygit2 / except ImportError` block above it is now unreachable, along with the three `PYGIT2_AVAILABLE` fallbacks it feeds. Removing those means rewriting two tests that flip the flag — happy to do it in this PR if you want it done here.

taf/lfs.py
dgreisen-llm: Done — removed, and I made a pass for the same pattern elsewhere: comments and docstrings that explained a change, a bug or a test rather than the code.

taf/lfs.py
dgreisen-llm: Done — `get_git_lfs_executable`, and the same `get_` prefix on the other accessors.

taf/lfs.py
dgreisen-llm: Done — the invocation is behind `get_git_config()` now, so the filter logic reads at one level. Why it is still a direct call rather than `taf.utils.run`: see my reply on the `lfs_utils.py` thread.

taf/lfs.py
dgreisen-llm: Done — the messages now say what happened and what to do ("left as a pointer … run 'git lfs pull'"), with the internal detail moved to debug level. One aggregate message also claimed content was unchanged on disk, which was wrong for a smudge; that is fixed too.

setup.py
dgreisen-llm: Done — removed. The rationale for the pin is in the commit message and the CHANGELOG.

setup.py
dgreisen-llm: Done — classifiers are 3.10–3.12 and `python_requires` is `>=3.10`, matching the CI matrix.

taf/lfs.py
dgreisen-llm: Good catch, and it was worse than broad — gone entirely. Nothing in the branch matches on a pointer prefix now; the tests ask `git lfs pointer --check --stdin`, which tells a real pointer from content that merely starts with the same line.

taf/tests/test_updater/test_lfs/test_lfs.py
dgreisen-llm: Done, and where `GitRepository` already exposes the operation I used that rather than `_git`: `commit_empty`, `merge_branch`, `commit`, `top_commit_of_branch`, `init_repo`. `_git` covers what it does not expose - reading and writing config, `lfs install`, `lfs track`, `status --porcelain`, `rev-list`. Two calls stay on `taf.utils.run`, both `git lfs pointer --check --stdin`, which needs bytes on stdin and has no repository to hang off.

taf/exceptions.py
dgreisen-llm: `GitLFSError` inherits `GitError` now, with the message first so the raise sites read as before, and a new test pins that a failed checkout is catchable as `GitError`.

On the handling, I audited every broad handler wrapping the six pygit2 entry points the filter runs in (`checkout_branch`, `checkout_paths`, `branch_off_commit`, `create_and_checkout_branch`, `merge_branch`, `something_to_commit`). Seven sites. Five already catch `Exception`, so the base class changes nothing there. Two catch `GitError` and would now turn a filter failure into a different message: `lifecycle_handlers.py:192` reports it as "does not contain any scripts", and `updater_pipeline.py:2300` as "Could not checkout branch ... update manually".

I did not add `except GitLFSError: raise` at either, because I could not reach the condition in a test without mocking: the update path I can drive fails earlier, in `git reset --hard`, where git-lfs' own message already reaches the user. Say the word and I will add both as defensive guards - I did not want to add a branch no test covers.

taf/lfs/lfs.py
```diff
@@ -0,0 +1,370 @@
+"""Make libgit2 checkouts honor Git LFS, by delegating to the git-lfs binary.
```
dgreisen-llm: Done - `taf/lfs/` holding `lfs.py`, `lfs_process.py` and `lfs_protocol.py` beside an empty `__init__.py`, the same layout as `taf/yubikey/`.

`taf/tests/test_lfs_protocol.py` stayed where it is. Moving it under the LFS test package would put the pkt-line unit tests, which need no binary and no repository, behind that package's autouse LFS-config fixture.

dgreisen-llm:
Split into a stack, since you asked for another pass over the LFS internals and 2,700 lines is a lot to ask of a second reviewer.

- #801 test: a deterministic git environment for the suite (+82)
- #802 build: a pygit2 with the filter API (+4)
- #803 feat: the pkt-line protocol and the long-running git-lfs process (+674)
- #766 feat: the filter itself, its wiring and the LFS tests (+1953) — this PR

#766 keeps the filter, so your comments on `taf/lfs.py`, `taf/git.py` and `taf/exceptions.py` are still against the code they were written on. The ones on `setup.py` moved to #802 and the one on `taf/tests/conftest.py` to #801; both are answered there in the descriptions.

Also rebased onto master. Two of this branch's commits are gone because master arrived at the same place independently: the securesystemslib cap, and the `get_commit_date` timezone fix — I kept the test for that one, since master has none.

PR #769, opened 2026-08-19
feat: safe_cleanup with graceful shutdown

MarkoSekulic:

added safe_cleanup decorator in taf.tools.cli for handling Graceful Shutdown (and final cleanup)
used said decorator on Pipeline.run()
moved remove_temp_repositories from handle_error to safe_cleanup
attempted to fix bugs with cleanup on updates

### DIFF
--- a/taf/tests/test_updater/test_pipeline_cleanup.py
+++ b/taf/tests/test_updater/test_pipeline_cleanup.py
@@ -1,0 +1,87 @@
+import pytest
+import signal
+from unittest.mock import Mock, patch
+from taf.updater.lifecycle_handlers import Event
+from taf.tools.cli import safe_cleanup
+from taf.updater.updater_pipeline import AuthenticationRepositoryUpdatePipeline
+
+
+@pytest.fixture
+def dummy_pipeline():
+    """Mock pipeline with required attributes for testing the decorator."""
+    pipeline = Mock(spec=AuthenticationRepositoryUpdatePipeline)
+    pipeline.only_validate = False
+    pipeline.state = Mock()
+    pipeline.state.existing_repo = False
+    pipeline.state.users_auth_repo = "some_repo"
+    pipeline.state.event = None
+    pipeline.remove_temp_repositories = Mock()
+    return pipeline
+
+
+def test_pipeline_cleanup_calls_cleanup_on_success(dummy_pipeline):
+    """On normal completion, cleanup is called once."""
+
+    @safe_cleanup
+    def dummy_method(self):
+        return "success"
+
+    result = dummy_method(dummy_pipeline)
+    assert result == "success"
+    dummy_pipeline.remove_temp_repositories.assert_called_once()
+
+
+def test_pipeline_cleanup_handles_keyboard_interrupt(dummy_pipeline):
+    """On Ctrl+C, cleanup is called and state is set to FAILED."""
+
+    @safe_cleanup
+    def dummy_method(self):
+        raise KeyboardInterrupt("Simulated Ctrl+C")
+
+    with pytest.raises(KeyboardInterrupt):
+        dummy_method(dummy_pipeline)
+
+    dummy_pipeline.remove_temp_repositories.assert_called_once()
+    assert dummy_pipeline.state.event == Event.FAILED
+
+
+def test_pipeline_cleanup_handles_normal_exception(dummy_pipeline):
+    """On a normal exception, cleanup is still called."""
+
+    @safe_cleanup
+    def dummy_method(self):
+        raise ValueError("Something went wrong")
+
+    with pytest.raises(ValueError):
+        dummy_method(dummy_pipeline)
+
+    dummy_pipeline.remove_temp_repositories.assert_called_once()
+
+
+@patch("signal.signal")
+def test_pipeline_cleanup_registers_and_restores_signal_handlers(
+    mock_signal, dummy_pipeline
+):
+    """Ensure SIGINT and SIGTERM are overridden temporarily."""
+    original_sigint = Mock()
+    original_sigterm = Mock()
+    mock_signal.side_effect = [
+        original_sigint,
+        original_sigterm,
+        original_sigint,
+        original_sigterm,
+    ]
+
+    @safe_cleanup
+    def dummy_method(self):
+        return "ok"
+
+    dummy_method(dummy_pipeline)
+
+    assert mock_signal.call_count == 4
+    # Setting handlers
+    assert mock_signal.call_args_list[0][0][0] == signal.SIGINT
+    assert mock_signal.call_args_list[1][0][0] == signal.SIGTERM
+    # Restoring handlers
+    assert mock_signal.call_args_list[2][0][0] == signal.SIGINT
+    assert mock_signal.call_args_list[3][0][0] == signal.SIGTERM
--- a/taf/tests/test_updater/test_pipeline_cleanup.py
+++ b/taf/tests/test_updater/test_pipeline_cleanup.py
@@ -1,3 +1,4 @@
+import types
 import pytest
 import signal
 from unittest.mock import Mock, patch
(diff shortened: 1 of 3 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -2043,8 +2050,12 @@ def remove_temp_repositories(self):
             # renaming/removing the temp directory on Windows
             for repo in self.state.temp_target_repositories.values():
                 repo.cleanup()
-            self.state.update_handler.cleanup()
+            # self.state.update_handler.cleanup() # This doesn't exist and is unnecesary
```
renatav, line 2053: Having looked at the broader code, I do think this exists, in `updater/handlers.py`. If it was `None` that's a different problem. It also does some important cleanup. Could you please revert this change and look again why that was failing? Perhaps there is a different issue we need to resolve

renatav (commented):
I'll most likely leave another comment, but wanted to flag this first. Sorry for not looking into it sooner

taf/tools/cli/__init__.py
```diff
@@ -156,3 +160,93 @@ def common_repo_edit_options(func):
...
+    Decorator that handles gracefull shutdown (SIGINT and SIGTERM)
+    applicable for:
+    Pipeline
+    any with cleanup()
+    """
+
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+
+        is_pipeline = "Pipeline" in self.__class__.__name__
```
renatav, line 176: This works, but is fragile` (any future class with "Pipeline" in the name but different shape would crash here)

Since the docstring already says "any with cleanup()," could we make Pipeline actually fit that contract instead of special-casing it? Then safe_cleanup just does `hasattr(self, "cleanup")`/ `hasattr(self, "on_interrupt")` everywhere.

taf/tools/cli/__init__.py
```diff
@@ -156,3 +160,93 @@ def common_repo_edit_options(func):
...
+                        pass
+                if original_sigterm is not None:
+                    try:
+                        signal.signal(signal.SIGTERM, original_sigterm)
+                    except Exception:
+                        pass
+
+            # Cleanup lives here (and ONLY here)
+            if is_pipeline:
+                self.remove_temp_repositories(final_cleanup=True)
```
renatav, line 245: I'd remove this from  `safe_cleanup`'s finally block add it to `Pipeline.run()`'s own finally:

finally:
    self.set_output()
    self.remove_temp_repositories(final_cleanup=True)

Finally blocks in Python run on every way out of a frame including a KeyboardInterrupt. I think that would be

taf/tests/test_updater/test_pipeline_cleanup.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.fixture
+def dummy_pipeline():
+    """Mock pipeline with required attributes for testing the decorator."""
+    pipeline = Mock(spec=AuthenticationRepositoryUpdatePipeline)
+    pipeline.only_validate = False
+    pipeline.state = Mock()
+    pipeline.state.existing_repo = False
+    pipeline.state.users_auth_repo = "some_repo"
+    pipeline.state.event = None
+    pipeline.remove_temp_repositories = Mock()
```
renatav, line 18: Since this is being mocked here, we aren't really testing `remove_temp_repositories` that was updated in this PR. Is any of the existing test already invoking this part of the pipeline? Could you verify?

taf/tools/cli/__init__.py
```diff
@@ -156,3 +160,93 @@ def common_repo_edit_options(func):
...
+
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+
+        is_pipeline = "Pipeline" in self.__class__.__name__
+
+        has_cleanup = hasattr(self, "cleanup") and callable(self.cleanup)
+
+        # Signal handler only mutates state; deletion is delegated to finally.
+        def _signal_handler(signum, frame):
```
renatav, line 181: This FAILED-state condition on interrupt is narrower than handle_error's unconditional version, so Ctrl+C during a validate-only/existing-repo/early run doesn't get marked FAILED like a crash would.

I think this handler could shrink to just:

def _signal_handler(signum, frame):
    raise KeyboardInterrupt(f"Received signal {signum}")

Could you test?

taf/tools/cli/__init__.py
```diff
@@ -156,3 +160,93 @@ def common_repo_edit_options(func):
...
+                signal.signal(signal.SIGTERM, _signal_handler)
+            except (OSError, ValueError):
+                pass  # Windows: SIGTERM cannot be caught
+
+        try:
+            return method(self, *args, **kwargs)
+
+        except KeyboardInterrupt:
+            # Set state for non-KeyboardInterrupt exceptions (ValueError, SystemExit, etc.)
+            if is_pipeline:
```
renatav, line 218: This duplicates logic that Pipeline.run()'s own except KeyboardInterrupt  If we drop the condition there (per the earlier comment), this whole block becomes unnecessary too. Tthis except wouldn't need to set state.event at all, just let it propagate.

taf/tools/cli/__init__.py
MarkoSekulic: I agree.

taf/tools/cli/__init__.py
MarkoSekulic: it could go into Pipeline's cleanup method that could be introduced, this is according to the first comment

taf/tools/cli/__init__.py
```diff
@@ -156,3 +159,71 @@ def common_repo_edit_options(func):
...
+        has_on_interrupt = hasattr(self, "on_interrupt") and callable(self.on_interrupt)
+        has_after_interrupt = hasattr(self, "after_interrupt") and callable(
+            self.after_interrupt
+        )
+
+        # Signal handler only mutates state; deletion is delegated to finally.
+        def _signal_handler(signum, frame):
+            if has_on_interrupt:
+                self.on_interrupt()
+            raise KeyboardInterrupt(f"Received signal {signum}")
```
renatav, line 183: The interrupt message showed raw signal info (Received signal 2), which isn't meaningful to taf's actual users as they're not developers. I suggest something like "The update was interrupted before it could finish.".

taf/tests/test_updater/test_pipeline_cleanup.py
```diff
@@ -0,0 +1,165 @@
+import types
+import pytest
+import signal
+from unittest.mock import Mock, patch
+from taf.updater.lifecycle_handlers import Event
+from taf.tools.cli import safe_cleanup
+from taf.updater.updater_pipeline import AuthenticationRepositoryUpdatePipeline
+
+
+@pytest.fixture
```
renatav, line 10: If thought a bit more about the tests and I think it's fine to leave some of this to manual testing.  We'd needs a real OS signal delivered at a precise moment while real blocking I/O is in progress. Any attempt would be timing-based and flaky rather than a reliable regression test.

taf/tests/test_updater/test_pipeline_cleanup.py
```diff
@@ -0,0 +1,165 @@
...
+
+@patch("signal.signal")
+def test_pipeline_on_interrupt_sets_failed_event_via_signal(
+    mock_signal, dummy_pipeline
+):
+    """A real interrupt (SIGINT/SIGTERM) fires the registered handler, which calls
+    on_interrupt and marks the update FAILED; cleanup still runs afterwards."""
+    captured_handlers = {}
+
+    def fake_signal(sig, handler):
```
renatav, line 93: This is the following lines appear in the next test too. Let's move it all to helper functions and reuse.

renatav (commented):
Minor comments and please update the changelog, we can then merge this in.

One thing worth mentioning, while testing this I hit a case where Ctrl+C didn't take effect right away during taf repo clone, and a second Ctrl+C surfaced an ugly raw traceback from Python's own thread shutdown. Several pipeline steps (clone_target_repositories_to_temp and others) clone target repos in parallel using `ThreadPoolExecutor`. When you press Ctrl+C while target repos are being cloned, the clones that are already running don't stop right away. Python can start several clones at once using background threads, but once one of those threads is in the middle of actually copying a repo, there's no way to force it to stop from outside. Python just doesn't have a way to kill a thread that's busy. So this traceback is out of scope of this work

taf/updater/updater_pipeline.py
```diff
@@ -2058,8 +2063,13 @@ def remove_temp_repositories(self):
             # renaming/removing the temp directory on Windows
             for repo in self.state.temp_target_repositories.values():
                 repo.cleanup()
-            self.state.update_handler.cleanup()
+            if self.state.update_handler is not None:
+                self.state.update_handler.cleanup()  # TODO: This should not be None.
```
renatav, line 2067: I've looked into this and in some cases this can be `None`, it's not a bug. Adding this `if` here was exactly the right fix, so let's just remove the `TODO`

renatav (approved):

renatav:
My only remaining comment is to update the changelog, but I can do that later too, not a blocker

PR #770, opened 2026-08-20
fix(yubikey): allow --key-pin with multiple YubiKeys inserted

n-dusan:
Previously any --key-pin usage hard-errored as soon as more than one YubiKey was inserted, since the pin could only be bound to "the" single detected key. Instead, narrow the inserted YubiKeys down to those that are valid signing keys for the roles being loaded; if exactly one candidate remains, bind the pin to it. If the result is still ambiguous (zero or multiple candidates), fall back to the existing interactive PIN prompt instead of raising.

### DIFF
--- a/taf/api/api_workflow.py
+++ b/taf/api/api_workflow.py
@@ -198,12 +198,12 @@ def _resolve_key_id_pins(
     Given a single --key-pin and the roles about to be loaded, figure out which
     of the currently inserted YubiKeys the pin belongs to.
 
-    A YubiKey is a candidate if it is a valid signing key for at least one of
-    `all_roles`. If exactly one candidate is found, the pin is bound to its
-    keyid. Otherwise (no candidates, or more than one), the pin is not bound
-    to anything and a warning is logged - the caller falls back to the normal
-    interactive PIN prompt (or PIN_<serial>/PIN_<key_name> env vars) instead
-    of erroring out.
+    A YubiKey is a candidate if any of its occupied PIV slots holds a valid
+    signing key for at least one of `all_roles`. If exactly one candidate is
+    found, the pin is bound to its keyid. Otherwise (no candidates, or more
+    than one), the pin is not bound to anything and a warning is logged - the
+    caller falls back to the normal interactive PIN prompt (or
+    PIN_<serial>/PIN_<key_name> env vars) instead of erroring out.
     """
     import taf.yubikey.yubikey as yk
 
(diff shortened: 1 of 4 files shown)


taf/api/api_workflow.py
```diff
@@ -204,6 +191,68 @@ def _setup_auth_repo_and_signers(
...
+    for serial_num in serial_nums:
+        try:
+            public_key = yk.get_piv_public_key_tuf(serial=serial_num)
+        except Exception as e:
+            taf_logger.debug(
+                f"Could not read public key of YubiKey serial={serial_num}: {e}"
+            )
+            continue
+
+        for role in all_roles:
```
renatav: I think we should already have some existing code that does this. Could we use `find_associated_roles_of_key(public_key)` ?

taf/api/api_workflow.py
```diff
@@ -204,6 +191,68 @@ def _setup_auth_repo_and_signers(
...
+    if len(serial_nums) == 0:
+        taf_logger.error("No Yubikeys found")
+        raise TAFError(
+            "Passed in a --key-pin but no YubiKeys inserted. Please insert a YubiKey and try again."
+        )
+
+    candidates = []
+    for serial_num in serial_nums:
+        try:
+            public_key = yk.get_piv_public_key_tuf(serial=serial_num)
```
renatav: Looks like `get_piv_public_key_tuf` only ever reads the `SIGNATURE` slot. We should look for these places that I failed to update and address this. Unless you want to release this as I point release without the slots support, in which case we should create an issue

taf/api/api_workflow.py
```diff
@@ -204,6 +191,68 @@ def _setup_auth_repo_and_signers(
...
+                    break
+            except TAFError:
+                continue
+
+    if len(candidates) == 1:
+        serial_num, public_key = candidates[0]
+        keyid = _get_legacy_keyid(public_key)
+        return {keyid: key_pin}
+
+    candidate_serials = [serial_num for serial_num, _ in candidates]
```
renatav: Could you check whether serials are numbers or strings? Think this will raise an error if they are ints.

taf/tests/test_api/test_key_id_pins.py
```diff
@@ -0,0 +1,104 @@
...
+
+import pytest
+
+import taf.api.api_workflow as api_workflow
+import taf.yubikey.yubikey as yk
+from taf.exceptions import TAFError
+from taf.tuf.keys import _get_legacy_keyid, load_signer_from_file
+
+
+def _real_public_key(keystore_delegations, key_name):
```
renatav: This could be a generic fixture in `conftest`

taf/api/api_workflow.py
```diff
@@ -204,6 +191,69 @@ def _setup_auth_repo_and_signers(
...
+    import taf.yubikey.yubikey as yk
+
+    serial_nums = yk.get_serial_nums()
+    if len(serial_nums) == 0:
+        taf_logger.error("No Yubikeys found")
+        raise TAFError(
+            "Passed in a --key-pin but no YubiKeys inserted. Please insert a YubiKey and try again."
+        )
+
+    roles_of_interest = set(all_roles)
```
renatav, line 219: Could you test the case when when one inserted YubiKey has two valid slots? Will candidates become length 2, and key-pin treated as ambiguous?

n-dusan: Good catch, resolved.

renatav (approved):

PR #774, opened 2026-08-27
fix(updater): resolve allow-unauthenticated-commits per auth commit

### LINKED ISSUE #768: Support changing the `allow-unauthenticated-commits` flag for a repo
Add tests and make sure the repos continue being valid after the flag is removed

n-dusan:
The updater resolved allow-unauthenticated-commits using only the newest repositories.json in the validated range, so flipping the flag from true to false retroactively invalidated unauthenticated commits that had already been signed over while it was still true, breaking every fresh clone or full-history validation of the repository.

Resolve the flag at the authentication commit that recorded each target advance instead, so older history stays valid and the change only applies going forward from the commit that disallows it.

Closes #768

### DIFF
--- a/taf/updater/updater_pipeline.py
+++ b/taf/updater/updater_pipeline.py
@@ -459,6 +459,7 @@ class AuthenticationRepositoryUpdatePipeline(Pipeline):
         self._output = None
         self.local_repos_consistent = True
         self.repos_synced_with_remote = False
+        self._repositories_json_cache: Dict[Commitish, Optional[Dict]] = {}
 
     @property
     def output(self):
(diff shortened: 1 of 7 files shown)


taf/updater/updater_pipeline.py
```diff
@@ -1857,6 +1858,48 @@ def validate_target_repositories(self):
             self.state.event = Event.FAILED
             return UpdateStatus.FAILED
 
+    def _get_repositories_json_at(
```
renatav: Even though we should consider fully reworking and removing repositoriesdb at some point, I think that currently we'd probably want to move some of this code to that module, as that would make it more consistent with the rest of the codebase. E.g. `get_repository_custom_at` and that keep `_is_unauthenticated_allowed_at` and simplify it.

renatav (approved):

PR #775, opened 2026-08-28
test(yubikey): add regression coverage for multi-YubiKey delegated-ro…

n-dusan:
…le signing

Add tests reproducing a reported bug: signing a delegated role while an unauthorized YubiKey is also inserted, in both insertion orders and across single- and multi-role signing commands. The scenario didn't reproduce as originally described, but reading the surrounding code turned up genuine order-dependent defects, fixed here:

- add_default_names_of_role zipped two independently-ordered lists, mis-pairing a key name with the wrong keyid when only some of a role's keys already had a name
- YubiKeyStore collided on key_name=None, letting two unnamed YubiKeys clobber each other's cached serial/public key/roles
- rejected YubiKeys were silently dropped instead of logged
- is_valid_metadata_yubikey(role) with no public_key only ever checked one device's SIGNATURE slot

### DIFF
--- a/taf/tests/tuf/test_keys/test_yubikey_delegated_signing.py
+++ b/taf/tests/tuf/test_keys/test_yubikey_delegated_signing.py
@@ -1,0 +1,222 @@
+"""Regression coverage for a reported multi-YubiKey bug: with two YubiKeys
+inserted, where only one of them holds a key with authority over a
+delegated role, signing was reported to fail depending on which key got
+checked first."""
+
+import pytest
+from tuf.api.metadata import Metadata
+from yubikit.piv import SLOT
+
+from taf.api.targets import register_target_files
+from taf.auth_repo import AuthenticationRepository
+from taf.tuf.keys import load_signer_from_file
+from taf.yubikey.yubikey_manager import PinManager
+import taf.yubikey.yubikey as yk
+
+
+def _pin_manager(*devices) -> PinManager:
+    pin_manager = PinManager()
+    for device in devices:
+        pin_manager.add_pin(device.serial, device.pin)
+    return pin_manager
+
+
+@pytest.mark.parametrize("authorized_inserted_first", [True, False])
+def test_read_and_check_yubikeys_skips_unauthorized_device_for_delegated_role(
+    make_fake_yubikey,
+    keystore_delegations,
+    create_delegated_auth_repo,
+    authorized_inserted_first,
+):
+    if authorized_inserted_first:
+        authorized = make_fake_yubikey(
+            "delegated_role1",
+            extra_slots={SLOT.AUTHENTICATION: "delegated_role2"},
+            keystore_path=keystore_delegations,
+        )
+        unauthorized = make_fake_yubikey("targets1", keystore_path=keystore_delegations)
+    else:
+        unauthorized = make_fake_yubikey("targets1", keystore_path=keystore_delegations)
+        authorized = make_fake_yubikey(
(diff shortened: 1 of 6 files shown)


taf/yubikey/yubikey.py
```diff
@@ -725,6 +728,13 @@ def _read_and_check_yubikeys(
             f"All YubiKeys inserted were already loaded for role='{role}'."
         )
 
+    if invalid_keys:
```
renatav, line 731: I haven't tested this, but based on the code it seems that, for each device, this checks only whether that device has a key valid for this specific role. If none of its occupied slots match, it's marked invalid, with being aware of whether it's valid for some other role being loaded in the same run. This might confuse the end users

taf/tests/tuf/test_keys/test_yubikey_delegated_signing.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+@pytest.mark.parametrize("authorized_inserted_first", [True, False])
+def test_read_and_check_yubikeys_skips_unauthorized_device_for_delegated_role(
+    make_fake_yubikey,
+    keystore_delegations,
+    create_delegated_auth_repo,
+    authorized_inserted_first,
+):
+    if authorized_inserted_first:
```
renatav: Nitpick: can be refactored to avoid some code repetition and can be extracted to a helper and reused by subsequent tests.

taf/tests/tuf/test_keys/test_yubikey_delegated_signing.py
```diff
@@ -0,0 +1,222 @@
...
+        unauthorized = make_fake_yubikey("targets1", keystore_path=keystore_delegations)
+    else:
+        unauthorized = make_fake_yubikey("targets1", keystore_path=keystore_delegations)
+        authorized = make_fake_yubikey(
+            "delegated_role1",
+            extra_slots={SLOT.AUTHENTICATION: "delegated_role2"},
+            keystore_path=keystore_delegations,
+        )
+
+    pin_manager = _pin_manager(authorized, unauthorized)
```
renatav: This and the following test seem to both contain the same setup code, so should reuse it. I'd also prefer to check whether we have existing setup functions that we can reuse, or move this setup code outside of these tests and make it reusable, as this diverges from our current standards. There's also a lot of low-level manipulation happening directly inside these tests, which makes them harder to read.

LLMs are really bad at properly organizing code and will copy-paste the same code unless you yell at them :D We might want to spend some times creating a proper `CLAUDE.md` and `AGENT.md` for this project.

taf/yubikey/yubikey.py
```diff
@@ -732,7 +732,8 @@ def _read_and_check_yubikeys(
         taf_logger.log(
             "NOTICE",
             f"YubiKey(s) with serial(s) {', '.join(str(s) for s in invalid_keys)} "
-            f"are not valid signing keys for role '{role}' and were skipped.",
+            f"do not hold a signing key for role '{role}' and were skipped for this "
+            "role - they may still be used to sign other roles in this run.",
```
renatav, line 736: This is better, but an ideal solution would probably be to acually wait until all roles are checked. That would probably be a bigger refactor, so if you agree, I'd create an issue to be address later

renatav (approved):

taf/yubikey/yubikey.py
n-dusan: Agreed. Let's create an issue

PR #777, opened 2026-09-02
feat(branch): add generic taf branch merge command

n-dusan:
CI's merge job reimplements, per branch type, work that belongs in taf: walking an authentication repository's unmerged commits by metadata version (git reachability alone can't tell what's merged, since auth commits are re-signed onto the default branch rather than fast-forwarded), fast-forwarding each target repository to its declared commit, and validating the result before push. Everything else - which repos participate, which branch each merges into, what date gates a commit - was hardcoded per branch type (publication vs rdf vs update) even though it's already present in the auth repo's signed target files: each target's commit, branch and gate date are recorded there per auth commit.

Add `taf branch merge`, driven by that data: a repo participates when its declared commit isn't yet reachable from its declared destination, so a repo untouched by a given commit (e.g. rdf not rebuilt) drops out on its own instead of needing a repo-type filter. The one thing not in that data - branch name patterns, capstone role policy - comes from a named MergePolicy, resolved from an explicit --config file, a root authentication repository's signed targets/merge-policies.json, or a built-in default, so per-repo behaviour can change without re-signing every partner repository. Update branches (single commit, no fan-out) fall out of the same model rather than needing a second command: their target commit is already on its destination by the time the branch is built, so nothing moves but the auth commit still gets signed.

branches.py gains select_branches, reusing the existing reachability-aware branch walk per group (e.g. one update branch per role) instead of a name-only sort, and find_unmerged_auth_commits drops its role parameter in favor of reading each commit's own changed metadata files.

### DIFF
--- a/taf/api/merge.py
+++ b/taf/api/merge.py
@@ -304,12 +304,19 @@ def _mirrored_repos(source_branch, merge_plan, repos_by_name, qualifying_steps, 
     left to the post-merge repository validation instead. Otherwise an uneven
     count is left in and surfaces as a validation error - the ordinary signal
     that a repo's branch lost commits to a force push.
+
+    Materializes a local `source_branch` for each mirrored repo (mirroring
+    `create_local_branch_from_remote_tracking` already used for `destination`)
+    so downstream git operations that only look at local branches - a caller's
+    own tooling, not anything this function needs itself - see it too.
     """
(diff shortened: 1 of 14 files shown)


n-dusan:
Closing in favor of #789 stack

